PR #1, opened 2022-03-21
Add check for compressed kernel

cfergeau:
The kernel passed to vfkit must be uncompressed, or the VM will silently
fail to start hanging at a black screen.
This commit checks if the kernel is compressed or not before attempting
to start the VM, and errors out if this is the case.

README.md has more details why an uncompressed kernel is required since
commit c92dd0416.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -2,8 +2,11 @@ package config
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/Code-Hex/vz"
+	"github.com/h2non/filetype"
+	"github.com/h2non/filetype/matchers"
 )
 
 type Bootloader struct {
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -27,7 +30,40 @@ func NewBootloader(vmlinuzPath, kernelCmdLine, initrdPath string) *Bootloader {
...
+	if err != nil {
+		return false, err
+	}
+	// uncompressed ARM64 kernels are matched as a MS executable, which is
+	// also an archive, so we need to special case it
+	if kind == matchers.TypeExe {
+		return false, nil
+	}
+
+	return filetype.IsArchive(buf), nil
```
anjannath, line 55: Based on the comment this `filetype.IsArchive()` is not strictly necessary, right? we can directly `return true, nil`

anjannath (approved):

pkg/config/config.go
cfergeau: Right, some left-over from the initial implementation. Could be useful if some day the file detection code no longer detects the kernel as `matchers.TypeExe`. Let me update this.

cfergeau:
Longer term, might be better to add our own pattern and match aarch64 kernel against it:
https://github.com/file/file/blob/25d5933179067a438c812637e230353e1bdae38d/magic/Magdir/linux#L252-L260

PR #4, opened 2022-07-21
Add virtiofs support

cfergeau:
This will be needed in order to get filesharing support in crc (`podman run --volume /hostdir:/containerdir`)


gbraad (approved):

cfergeau:
The latest force-push removes some unneeded downgrades in go.mod in the latest commit, no changes apart from that.

PR #5, opened 2022-07-22
Make use of github actions

cfergeau:
This fixes the default branch name (it's 'main' not 'master') so that ghactions are useful. This also switches to go 1.17, and tests builds on macOS 12.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -12,6 +12,7 @@ jobs:
       matrix:
         os:
         - macOS-11
+        - macOS-12
         go:
         - 1.17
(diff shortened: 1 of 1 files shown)


cfergeau:
The latest commit (artifact upload) depends on https://github.com/code-ready/vfkit/pull/3

anjannath (approved):

cfergeau:
Regarding the artifact upload, it's better to release the binary built on macOS11 as in my testing it was compatible with both macOS 11 and 12. The binary built on macOS 12 does not run on macOS 11.

PR #18, opened 2022-11-21
Add EFI support

cfergeau:
This PR adds support for EFI for linux virtual machines
https://developer.apple.com/documentation/virtualization/vzefibootloader?language=objc

This allows to boot directly from a disk image without the need for an external kernel/initrd.
This requires macOS 13.

This is done with a new --bootloader option which replaces the (now legacy)
--kernel/--initrd/--kernel-cmdline parameters.


pkg/client/cmdline.go
```diff
@@ -12,21 +12,32 @@ import (
...
+// linuxBootloader determines which kernel/initrd/kernel args to use when starting
+// the virtual machine.
+type linuxBootloader struct {
 	vmlinuzPath   string
 	kernelCmdLine string
 	initrdPath    string
 }
 
+// efiBootloader determines which kernel/initrd/kernel args to use when starting
+// the virtual machine.
```
praveenkumar, line 29: Looks like it is copy/paste from LinuxBootloader.

cmd/vfkit/main.go
```diff
@@ -25,54 +25,61 @@ import (
...
 	"github.com/crc-org/vfkit/pkg/vf"
 	"github.com/docker/go-units"
 	log "github.com/sirupsen/logrus"
 )
 
-type cmdlineOptions struct {
-	vcpus     uint
-	memoryMiB uint
+func newLegacyBootloader(opts *cmdline.Options) config.Bootloader {
+	if opts.VmlinuzPath == "" && opts.KernelCmdline == "" && opts.InitrdPath == "" {
```
praveenkumar, line 37: Shouldn't this with `||` instead `&&`, if anything of those options are empty string should be nil?

cfergeau: Below there is `cmd.MarkFlagsRequiredTogether("kernel", "initrd", "kernel-cmdline")` so if one is set, all must be set. The behaviour was the same before this commit, they were all required (even if that's not 100% correct, kernel should be mandatory and the other optional).
I would keep the && for now.

pkg/client/cmdline.go
cfergeau: Thanks, fixed.

cmd/vfkit/main.go
praveenkumar: Alright make sense.

praveenkumar (approved):

PR #20, opened 2022-12-16
Rearrange code in pkg/

cfergeau:
At the moment, vfkit has:
- pkg/client for client applications, this exposes go structures which are then converted to a vfkit commandline
- pkg/config which is used by vfkit itself, goes from commandline to go struct, which are then used to call into Code-Hex/vz API
- pkg/vf which has some virtio-vsock code using Code-Hex/vz API

The initial reason for the split between pkg/client and pkg/config was to make sure clients (crc) using vfkit do not need the Code-Hex/vz module, as this would mean only native builds are possible.
However, it comes with drawbacks as pkg/client and pkg/config duplicates quite a few go structures, and it's odd to have go->cmdline code in one package, and cmdline->go code in another.

This PR will merge pkg/client and pkg/config, and will move all the code using Code-Hex/vz to pkg/vf. Overall it's "just" some code movement, I tried to do things in small steps, so it's a bit of a lengthy PR.


anjannath (approved):

anjannath:
Built vfkit from the PR and tested with CRC and its working as expected

cfergeau:
Thanks for the testing!

PR #27, opened 2023-04-19
Improve Virtualization Framework error reporting

cfergeau:
This PR improves vfkit behaviour when there's an error from the Virtualization Framework.
Before this PR, some errors are ignored and never reported. After this PR, error reporting should be improved.
This also adds a convenient start-crc-bundle script, and fixes a doc typo.


cmd/vfkit/main.go
```diff
@@ -154,6 +156,13 @@ func runVirtualMachine(vmConfig *config.VirtualMachine) error {
 	log.Infof("waiting for VM to stop")
 	for {
 		err := waitForVMState(vm, vz.VirtualMachineStateStopped)
+		if err == vmStateTimeoutErr {
+			continue
+		}
+		if err != nil && errors.Is(err, vmStateTimeoutErr) {
```
anjannath, line 162: should this be `if err != nil && !error.Is(err, vmStateTimeoutErr)`?

cfergeau: ouch yes, thanks! Use of `errors.Is` was a last minute change, too rushed ^^

cfergeau: I changed the whole block doing the error checks, but I can get back to something similar to this if that's better.

PR #29, opened 2023-04-23
Add Restful service

baude:
add restful service to VFKit which defaults to `tcp://localhost:8081`.  The URL can be changed using the newly introduced --restful-uri option.  The schemes may be `tcp` or `unix` (unix domain socket).  The service for the unix domain implementation is incomplete and should be considered a TODO still.  users can also opt out of the service with a `--restful-uri none`.

the restful service now replies to three endpoints:

/vm/state GET which returns the state of the vm
/vm/state POST which allows for a state change
/vm/inspect GET which returns basic information about the virtual machine

error handling for the endpoints needs to be completed still.  these endpoints should return http response codes (like 200) based on the error handling.  this can be completed in a subsequent PR.

/version GET should also probably be added. I was unsure if we should return the version of vfkit (my first inclination) or some sort of API version.  maybe we should return both?

added tests for new content where applicable


pkg/config/state_change.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (vm *VirtualMachine) Resume() error {
+	logrus.Debug("resuming machine")
+	return vm.VzVM.Resume()
+}
+
+func (vm *VirtualMachine) Stop() error {
+	logrus.Debug("stopping machine")
+	return vm.VzVM.Stop()
```
cfergeau, line 35: `Stop()` is a  forceful stop of the VM. `RequestStop()` is the graceful stop.

pkg/rest/define/config.go
```diff
@@ -0,0 +1,32 @@
+package define
```
cfergeau, line 1: We call this `types` in crc, though if you are used to `define` this also works for me.

pkg/rest/define/config.go
```diff
@@ -0,0 +1,32 @@
...
+
+// StateChange is a string strong typing of values for changing
+// the state of a virtual machine
+type StateChange string
+
+const (
+	Resume    StateChange = "Resume"
+	Pause     StateChange = "Pause"
+	Stop      StateChange = "Stop"
+	ForceStop StateChange = "Hardstop"
```
cfergeau, line 31: Can the naming be consistent, `ForceStop` or `HardStop`

cfergeau (commented):
Here is a first round of comments, I need to take a closer look at the REST API.

pkg/config/state_change.go
cfergeau: One caveat with the `Stop` API is that it's only supported with macOS 12+, an error will be returned on macOS 11. This can be wrapped in a `CanStop()` call, which will always return false on macOS 11. It's ok to fix that later in the future, I'm mentioning it while I have it in mind.

cfergeau:
> /version GET should also probably be added. I was unsure if we should return the version of vfkit (my first inclination) or some sort of API version. maybe we should return both?

yeah vfkit version is good to have (possibly with git hash if that's easy to get). 
API version, if that's about the REST API, I'd keep this for later. I'm sure we'll want to bump the API some day, but will we do this through a `/version` entry point, through a `/v2` in the URL, .. I have no idea.

pkg/cmdline/cmdline.go
```diff
@@ -34,6 +43,26 @@ func AddFlags(cmd *cobra.Command, opts *Options) {
 	cmd.Flags().UintVarP(&opts.MemoryMiB, "memory", "m", 512, "virtual machine RAM size in mibibytes")
 
 	cmd.Flags().StringVarP(&opts.TimeSync, "timesync", "t", "", "sync guest time when host wakes up from sleep")
-
 	cmd.Flags().StringArrayVarP(&opts.Devices, "device", "d", []string{}, "devices")
+
+	cmd.Flags().StringVar(&opts.RestfulURI, "restful-uri", DefaultRestfulURI, "URI address for RestFul services")
```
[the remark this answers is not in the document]
baude: that's cool by me

doc/usage.md
```diff
@@ -203,3 +215,31 @@ The share can be mounted in the guest with `mount -t virtio-fs vfkitTag /mnt`, w
...
+POST `/vm/state` {"new_state": "new value"}
+
+Response: http 200
+
+### Inspect VM
+
+Get description of the virtual machine
+
+GET `/vm/inspect`
+Response: { "cpus": uint, "memory": uint64, "devices": []config.VirtIODevice }
```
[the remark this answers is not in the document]
baude: Ill be adding that in the next PR for sure.

pkg/rest/rest.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+	}
+}
+
+// setVMState requests a state change on a virtual machine.  At this time only
+// the following states are valid:
+// Pause - pause a running machine
+// Resume - resume a paused machine
+// Stop - stops a running machine
+// Forcestop - forceably stops a running machine
```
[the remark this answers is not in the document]
baude: fixed

pkg/config/state_change.go
baude: fixed

pkg/cmdline/cmdline.go
```diff
@@ -36,8 +43,26 @@ func AddFlags(cmd *cobra.Command, opts *Options) {
 	cmd.Flags().UintVarP(&opts.MemoryMiB, "memory", "m", 512, "virtual machine RAM size in mibibytes")
 
 	cmd.Flags().StringVarP(&opts.TimeSync, "timesync", "t", "", "sync guest time when host wakes up from sleep")
-
 	cmd.Flags().StringArrayVarP(&opts.Devices, "device", "d", []string{}, "devices")
 
 	cmd.Flags().StringVar(&opts.LogLevel, "log-level", "", "set log level")
+	cmd.Flags().StringVar(&opts.RestfulURI, "restful-uri", DefaultRestfulURI, "URI address for RestFul services")
```
cfergeau, line 49: Any reason for using `none://` instead of `""` as the default?

cmd/vfkit/main.go
```diff
@@ -110,18 +112,41 @@ func waitForVMState(vm *vz.VirtualMachine, state vz.VirtualMachineState) error {
...
+		return err
+	}
 	vm, err := vz.NewVirtualMachine(vzVMConfig)
 	if err != nil {
 		return err
 	}
 
-	err = vm.Start()
+	// Do not enable the rests server if user sets scheme to None
+	if scheme != rest.None {
```
cfergeau, line 136: I'd prefer to just have: 
```
    if opts.RestfulURI != "" { // or cmdline.DefaultRestURI instead of ""
        srv, err := rest.NewServer(..., opts.RestfulURI)
        // error handling
        go func() {
            if err := srv.Start(); err != nil {
				log.Error(err)
	    }
        }
````
This way we don't need to have the URL parsing (`ParseRestfulURI`, `ToRestScheme`) in `runVFKit`

File: pkg/rest/rest.go [hunk 9]
```diff
@@ -0,0 +1,141 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/crc-org/vfkit/pkg/rest/define"
+	"github.com/gorilla/mux"
```
cfergeau, line 12: For the record (we discussed this privately), https://github.com/gorilla/mux says "The Gorilla project has been archived, and is no longer under active maintainenance", we'll need to move to some other project.

pkg/rest/rest.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// inspect returns information about the virtual machine like hw resources
+// and devices
+func (vm *VzVirtualMachine) inspect() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		ii := define.InspectResponse{
+			// TODO complete me
+			CPUs:   1,
+			Memory: 2048,
```
cfergeau, line 55: You'll need a `config.VirtualMachine` to easily fill this.

pkg/rest/rest.go
```diff
@@ -0,0 +1,141 @@
...
+	r := mux.NewRouter()
+	s := VFKitService{
+		router: r,
+		Host:   uri.Host,
+		Scheme: scheme,
+	}
+	// Handlers for the restful service.  This is where endpoints are defined.
+	r.HandleFunc("/vm/state", vm.getVMState()).Methods(http.MethodGet)
+	r.HandleFunc("/vm/state", vm.setVMState()).Methods(http.MethodPost)
+	r.HandleFunc("/vm/inspect", vm.inspect()).Methods(http.MethodGet)
```
cfergeau, line 44: I don't expect vfkit and vz code to be fully thread safe, I don't know if the handlers are called in their own go routines, or if gorilla/its replacement run all of them in the same go routine, ...

cfergeau (commented):
Overall looks good! A few comments up for discussion

pkg/rest/rest.go
```diff
@@ -0,0 +1,138 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/crc-org/vfkit/pkg/rest/define"
+	"github.com/gin-gonic/gin"
```
cfergeau, line 11: The gin changes look fine.

pkg/cmdline/cmdline.go
baude: i prefer this approach because the default IS none and this saves us a conditional check later ...

pkg/rest/rest.go
```diff
@@ -0,0 +1,138 @@
...
+	Path   string
+	Port   int
+	router *gin.Engine
+	Scheme ServiceScheme
+}
+
+// Start initiates the already configured restful service
+func (v *VFKitService) Start() error {
+	logrus.Debugf("starting rest service on %s", v.Host)
+	return http.ListenAndServe(v.Host, v.router)
```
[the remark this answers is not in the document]
baude: yup, ill add the UDS after i have a decent proof of concept working

baude:
I have updated the PR per your wishes

PR #34, opened 2023-04-27
Remove whitespace and indents from OWNERS file

baude:


baude:
looks like i will have to rebase this one after #33 merges

PR #43, opened 2023-06-02
custom json methods for virtual machine

baude:
Add custom JSON methods for Marshal and Unmarshal so other projects can write a configuration file based on virtual machine.  the key issue is the use of interfaces, for example devices, in virtual machine.  When using JSON to marshal and unmarshal, we need to tag what kind of device it is so the process can tag and ultimately re-identify the proper device.


pkg/config/json.go
```diff
@@ -0,0 +1,237 @@
...
+		Kind vmComponentKind `json:"kind"`
+		LinuxBootloader
+	}
+	return json.Marshal(blWithKind{
+		Kind:            linuxBootloader,
+		LinuxBootloader: *bootloader,
+	})
+}
+
+func (dev *VirtioNet) MarshalJSON() ([]byte, error) {
```
jakecorrenti, line 173: Will need to add functions for `VirtioInput` and `VirtioGPU`

jakecorrenti:
@cfergeau friendly ping

baude:
rebased and updated for virtioinput and virtiogpu

@jakecorrenti ptal

jakecorrenti:
One small change, but lgtm otherwise

cfergeau:
Closing this in favour of https://github.com/crc-org/vfkit/pull/46 where I did some follow-up work.

PR #44, opened 2023-06-06
Add support for a GUI, virtio-input and virtio-gpu devices

jakecorrenti:
Adds command line support for the user to add virtio-gpu and virtio-input devices which also ties into having a good experience with the newly added gui compatability.

Adds command line support for the user to start a graphical application window when their virtual machine starts through the `--gui` flag (assuming the user added a virtio-gpu device as well).

Adds some mutual exclusion during creation of the virtual machine configuration. Without calling `runtime.LockOSThread()`, starting the virtual machine subsequently starting a graphic application window would result in a SIGILL error. Alongside calling `runtime.LockOSThread()`, parallelized the running of the graphic application window and checking the status of the virtual machine. Otherwise, the two wouldn't be able to run simultaneously.

These new features can be testing using a command similar to this:

```bash
$ ./out/vfkit --cpus 2 --memory 2048 \
--bootloader efi,variable-store=/Users/jakecorrenti/efi-variable-store,create \
--device virtio-serial,stdio \
--device virtio-fs,sharedDir=/Users/jakecorrenti,mountTag=dir0 \
--device virtio-blk,path=/Users/jakecorrenti/Downloads/Fedora-Server-dvd-x86_64-38-1.6.iso \
--device virtio-net,nat,mac=72:20:43:d4:38:62 --device virtio-input,keyboard --device virtio-input,pointing --device virtio-gpu --gui
```

### DIFF
--- a/cmd/vfkit/main.go
+++ b/cmd/vfkit/main.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"runtime"
 	"syscall"
 	"time"
 
@@ -113,10 +114,18 @@ func waitForVMState(vm *vz.VirtualMachine, state vz.VirtualMachineState) error {
 }
 
 func runVFKit(vmConfig *config.VirtualMachine, opts *cmdline.Options) error {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
 	vzVMConfig, err := vf.ToVzVirtualMachineConfig(vmConfig)
 	if err != nil {
 		return err
 	}
+
+	gpuDevs := vmConfig.VirtioGPUDevices()
+	if opts.UseGUI && len(gpuDevs) > 0 {
+		gpuDevs[0].UsesGUI = true
+	}
+
 	vm, err := vz.NewVirtualMachine(vzVMConfig)
 	if err != nil {
 		return err
@@ -168,18 +177,35 @@ func runVirtualMachine(vmConfig *config.VirtualMachine, vm *vz.VirtualMachine) e
 	}
 
 	log.Infof("waiting for VM to stop")
-	for {
-		err := waitForVMState(vm, vz.VirtualMachineStateStopped)
-		if err == nil {
-			log.Infof("VM is stopped")
+
+	errCh := make(chan error, 1)
+	go func() {
+		for {
+			err := waitForVMState(vm, vz.VirtualMachineStateStopped)
+			if err == nil {
+				log.Infof("VM is stopped")
+				errCh <- nil
+				return
+			}
+			if !errors.Is(err, errVMStateTimeout) {
+				errCh <- fmt.Errorf("virtualization error: %v", err)
+				return
+			}
+			// errVMStateTimeout -> keep looping
+		}
+	}()
+
+	for _, gpuDev := range vmConfig.VirtioGPUDevices() {
+		if gpuDev.UsesGUI {
+			runtime.LockOSThread()
+			err := vm.StartGraphicApplication(float64(gpuDev.Height), float64(gpuDev.Width))
+			runtime.UnlockOSThread()
+			if err != nil {
+				return err
+			}
 			break
 		}
-		if !errors.Is(err, errVMStateTimeout) {
-			log.Infof("virtualization error: %v", err)
-			return err
-		}
-		// errVMStateTimeout -> keep looping
 	}
 
-	return nil
+	return <-errCh
 }
--- a/pkg/config/virtio.go
+++ b/pkg/config/virtio.go
@@ -11,6 +11,37 @@ import (
 // The VirtioDevice interface is an interface which is implemented by all virtio devices.
 type VirtioDevice VMComponent
 
+const (
+	// Possible values for VirtioInput.InputType
+	VirtioInputPointingDevice = "pointing"
+	VirtioInputKeyboardDevice = "keyboard"
+
+	// Options for VirtioGPUResolution
+	VirtioGPUResolutionHeight = "height"
+	VirtioGPUResolutionWidth  = "width"
+
+	// Default VirtioGPU Resolution
+	defaultVirtioGPUResolutionHeight = 800
+	defaultVirtioGPUResolutionWidth  = 600
+)
+
+// VirtioInput configures an input device, such as a keyboard or pointing device
+// (mouse) that the virtual machine can use
+type VirtioInput struct {
+	InputType string `json:"inputType"` // currently supports "pointing" and "keyboard" input types
+}
+
+type VirtioGPUResolution struct {
+	Height int `json:"height"`
+	Width  int `json:"width"`
+}
+
+// VirtioGPU configures a GPU device, such as the host computer's display
+type VirtioGPU struct {
+	UsesGUI bool `json:"usesGUI"`
+	VirtioGPUResolution
+}
+
 // VirtioVsock configures of a virtio-vsock device allowing 2-way communication
 // between the host and the virtual machine type
 type VirtioVsock struct {
@@ -114,6 +145,10 @@ func deviceFromCmdLine(deviceOpts string) (VirtioDevice, error) {
 		dev = &VirtioVsock{}
 	case "usb-mass-storage":
 		dev = usbMassStorageNewEmpty()
+	case "virtio-input":
+		dev = &VirtioInput{}
+	case "virtio-gpu":
+		dev = &VirtioGPU{}
 	default:
 		return nil, fmt.Errorf("unknown device type: %s", opts[0])
 	}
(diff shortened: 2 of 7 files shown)


cmd/vfkit/main.go
```diff
@@ -168,18 +184,35 @@ func runVirtualMachine(vmConfig *config.VirtualMachine, vm *vz.VirtualMachine) e
...
 		}
-		if !errors.Is(err, errVMStateTimeout) {
-			log.Infof("virtualization error: %v", err)
-			return err
+	}()
+
+	for _, gpuDev := range vmConfig.VirtioGPUDevices() {
+		if gpuDev.UsesGUI {
+			runtime.LockOSThread()
+			err := vm.StartGraphicApplication(900, 600)
```
gbraad, line 208: are these the dimensions?

gbraad:
ok-to-test

pkg/config/virtio.go
```diff
@@ -11,6 +11,27 @@ import (
 // The VirtioDevice interface is an interface which is implemented by all virtio devices.
 type VirtioDevice VMComponent
 
+const (
```
baude, line 14: maybe it makes sense to break this group of devices into something like graphics.go or display.go.  Dealer's choice.

pkg/config/virtio.go
```diff
@@ -181,6 +206,109 @@ func (dev *VirtioSerial) FromOptions(options []option) error {
...
+
+func (dev *VirtioGPU) FromOptions(options []option) error {
+	for _, option := range options {
+		if option.key == VirtioGPUDisplayDevice {
+			if option.value != "" {
+				return fmt.Errorf(fmt.Sprintf("Unexpected value for virtio-gpu %s option: %s", option.key, option.value))
+			}
+			dev.Device = option.key
+		} else {
+			return fmt.Errorf("Unknown option for virtio-gpu devices: %s", option.key)
```
baude, line 305: consider moving this "up" as an if-condition and inverting the condition.   if option.key != VirtGPUDisplaDevice ...

pkg/config/virtio.go
```diff
@@ -181,6 +206,109 @@ func (dev *VirtioSerial) FromOptions(options []option) error {
...
+	if err := dev.validate(); err != nil {
+		return nil, err
+	}
+
+	return []string{"--device", fmt.Sprintf("virtio-gpu,%s", dev.Device)}, nil
+}
+
+func (dev *VirtioGPU) FromOptions(options []option) error {
+	for _, option := range options {
+		if option.key == VirtioGPUDisplayDevice {
```
baude, line 299: if you move up the error condition below, you can remove this conditional.

pkg/vf/virtio.go
```diff
@@ -53,6 +55,81 @@ func (dev *VirtioBlk) AddToVirtualMachineConfig(vmConfig *vzVirtualMachineConfig
...
+		return nil, fmt.Errorf("failed to initialize virtio graphic device: %w", err)
+	}
+	graphicsScanoutConfig, err := vz.NewVirtioGraphicsScanoutConfiguration(1920, 1200)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create graphics scanout: %w", err)
+	}
+	gpuDeviceConfig.SetScanouts(
+		graphicsScanoutConfig,
+	)
+	if err != nil {
```
baude, line 111: leftover?misplaced?

pkg/config/virtio.go
```diff
@@ -181,6 +206,109 @@ func (dev *VirtioSerial) FromOptions(options []option) error {
...
+		return nil, err
+	}
+	dev.(*VirtioGPU).UsesGUI = true
+	return dev, nil
+}
+
+func (dev *VirtioGPU) validate() error {
+	if dev.Device != VirtioGPUDisplayDevice {
+		return fmt.Errorf("Unknown option for virtio-gpu devices: %s", dev.Device)
+	} else if dev.Device == "" {
```
baude, line 282: optionally dump the extra else

pkg/config/virtio.go
```diff
@@ -181,6 +206,109 @@ func (dev *VirtioSerial) FromOptions(options []option) error {
...
+}
+
+func (dev *VirtioInput) FromOptions(options []option) error {
+	for _, option := range options {
+		if option.key == VirtioInputPointingDevice || option.key == VirtioInputKeyboardDevice {
+			if option.value != "" {
+				return fmt.Errorf(fmt.Sprintf("Unexpected value for virtio-input %s option: %s", option.key, option.value))
+			}
+			dev.InputType = option.key
+		} else {
```
baude, line 247: move "up" and invert the logic

pkg/config/virtio.go
```diff
@@ -181,6 +206,109 @@ func (dev *VirtioSerial) FromOptions(options []option) error {
...
+
+	return &VirtioInput{
+		InputType: inputType,
+	}, nil
+}
+
+func (dev *VirtioInput) validate() error {
+	if dev.InputType != VirtioInputPointingDevice && dev.InputType != VirtioInputKeyboardDevice {
+		return fmt.Errorf("Unknown option for virtio-input devices: %s", dev.InputType)
+	} else if dev.InputType == "" {
```
baude, line 225: optionally dump the else

pkg/config/virtio.go
```diff
@@ -11,6 +11,27 @@ import (
...
+
+// VirtioInput configures an input device, such as a keyboard or pointing device
+// (mouse) that the virtual machine can use
+type VirtioInput struct {
+	InputType string // currently supports "pointing" and "keyboard" input types
+}
+
+// VirtioGPU configures a GPU device, such as the host computer's display
+type VirtioGPU struct {
+	Device  string // currently supports "display" as the device
```
baude, line 31: please add json tags so podman can use them later.

cmd/vfkit/main.go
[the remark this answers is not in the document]
jakecorrenti: I was thinking option 2. Give the user the option to specify custom dimensions, but also default to something like 800x600

jakecorrenti: Fixed!

baude:
LGTM

pkg/config/virtio.go
```diff
@@ -181,6 +220,165 @@ func (dev *VirtioSerial) FromOptions(options []option) error {
...
+// The inputType parameter is the type of virtio-input device that will be added
+// to the machine.
+func VirtioInputNew(inputType string) (VirtioDevice, error) {
+	if inputType != VirtioInputPointingDevice && inputType != VirtioInputKeyboardDevice {
+		return nil, fmt.Errorf("Unknown option for virtio-input devices: %s", inputType)
+	}
+
+	return &VirtioInput{
+		InputType: inputType,
+	}, nil
```
cfergeau, line 233: This could be written as 
```
    dev := &VirtioInput{
		InputType: inputType,
	}
    if err := dev.validate(); err != nil {
        return nil, err
    }
    return dev, nil
```
to avoid repeating that same validation code multiple times.

pkg/config/virtio.go
```diff
@@ -181,6 +220,165 @@ func (dev *VirtioSerial) FromOptions(options []option) error {
...
+		Height: height,
+		Width:  width,
+	}
+	return dev, nil
+}
+
+// VirtioGPUNewWithGUIAndDimensions creates a new gpu device for the virtual
+// machine with the graphical user interface enabled and user defined
+// dimensions for its resolution.
+func VirtioGPUNewWithGUIAndDimensions(device string, height, width int) (VirtioDevice, error) {
```
cfergeau, line 313: I don't think you need all these constructors as only one or 2 are used ? I'd prefer to have just VirtioGPUNew(...) with a set of args which are mandatory, and a few `VirtioGPUSetXxx` if that's needed.

pkg/config/virtio.go
```diff
@@ -181,6 +220,165 @@ func (dev *VirtioSerial) FromOptions(options []option) error {
...
+			return fmt.Errorf(fmt.Sprintf("Unexpected value for virtio-input %s option: %s", option.key, option.value))
+		}
+		dev.InputType = option.key
+	}
+	return dev.validate()
+}
+
+// VirtioGPUNew creates a new gpu device for the virtual machine.
+// The device parameter is the virtio-gpu device that will be added to the machine
+func VirtioGPUNew(device string) (VirtioDevice, error) {
```
cfergeau, line 267: I'm not fully clear why this `device` string is required? This does not seem to be used in `Code-Hex/vz`, and it must be `display` in vfkit code?

cfergeau (commented):
Thanks a lot for your PR, this is going to be a useful addition to vfkit! 
I'm only half way through the review, but here are my first round of comments.

gbraad:
please be aware that the git title is too long and gets broken up. best to just write 'gui' instead of the full 'graphical user interface' as I care more about the addition of the last words: virtio-gpu devices

cfergeau:
> please be aware that the git title is too long and gets broken up. best to just write 'gui' instead of the full 'graphical user interface' as I care more about the addition of the last words: virtio-gpu devices

Since you bring this up, I would have preferred (at least) 3 separate commits, each doing one specific thing, "add virtio-input support", "add virtio-gpu support", "add gui support". This also helps with the shortlog length ;)

gbraad:
> I would have preferred (at least) 3 separate commits,

Didn't want to be so picky at first, but the description talks about the "I added", "I also added", which refers to these, right. Prefer not to see 'I' in a description either. More like 'This adds'

jakecorrenti:
> > I would have preferred (at least) 3 separate commits,
> 
> Didn't want to be so picky at first, but the description talks about the "I added", "I also added", which refers to these, right. Prefer not to see 'I' in a description either. More like 'This adds'
> 
> Let's say, for future commits to separate them?

Fixed. I fixed the PR info but forgot to actually change the commit. Also, feel free to be picky. I want this to be a quality contribution

gbraad (approved):

jakecorrenti:
> > please be aware that the git title is too long and gets broken up. best to just write 'gui' instead of the full 'graphical user interface' as I care more about the addition of the last words: virtio-gpu devices
> 
> Since you bring this up, I would have preferred (at least) 3 separate commits, each doing one specific thing, "add virtio-input support", "add virtio-gpu support", "add gui support". This also helps with the shortlog length ;)

I can definitely see this being better, I had the habit of squashing everything. I think it would be beneficial to have some sort of `CONTRIBUTING.md` file to have a location for contributors to see what you want for a PR so you don't have to make the same change requests.

gbraad:
we prefer small(er) commits as we have been in situations where bisect was required.

pkg/config/virtio.go
```diff
@@ -181,6 +220,122 @@ func (dev *VirtioSerial) FromOptions(options []option) error {
...
+// be displayed
+func VirtioGPUNew() (VirtioDevice, error) {
+	virtioGPU := &VirtioGPU{
+		Device:  VirtioGPUDisplayDevice,
+		UsesGUI: false,
+	}
+	virtioGPU.VirtioGPUResolution = VirtioGPUResolution{
+		Height: defaultVirtioGPUResolutionHeight,
+		Width:  defaultVirtioGPUResolutionWidth,
+	}
```
cfergeau, line 279: fwiw this can be done in a single statement as well:
```
virtioGPU := &VirtioGPU{
		Device:  VirtioGPUDisplayDevice,
		UsesGUI: false,
		VirtioGPUResolution: VirtioGPUResolution{
				Height: defaultVirtioGPUResolutionHeight,
				Width:  defaultVirtioGPUResolutionWidth,
		},
	}
```

I'm fine with your code, no need to change if you prefer your way.

cmd/vfkit/main.go
```diff
@@ -113,10 +117,22 @@ func waitForVMState(vm *vz.VirtualMachine, state vz.VirtualMachineState) error {
...
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
 	vzVMConfig, err := vf.ToVzVirtualMachineConfig(vmConfig)
 	if err != nil {
 		return err
 	}
+
+	if opts.UseGUI {
+		for _, gpuDev := range vmConfig.VirtioGPUDevices() {
+			if gpuDev.Device == config.VirtioGPUDisplayDevice {
```
cfergeau, line 129: I don't understand why this `Device` field is needed? It seems the code would work equally well if the field was removed?

pkg/config/virtio.go
```diff
@@ -181,6 +220,122 @@ func (dev *VirtioSerial) FromOptions(options []option) error {
...
+
+	return []string{"--device", fmt.Sprintf("virtio-gpu,%s,height=%d,width=%d", dev.Device, dev.Height, dev.Width)}, nil
+}
+
+func (dev *VirtioGPU) FromOptions(options []option) error {
+	for _, option := range options {
+		if option.key != VirtioGPUDisplayDevice && option.key != VirtioGPUResolutionWidth && option.key != VirtioGPUResolutionHeight {
+			return fmt.Errorf("Unknown option for virtio-gpu devices: %s", option.key)
+		}
```
cfergeau, line 307: This can be the `default:` case in the switch/case below.

pkg/config/virtio.go
```diff
@@ -181,6 +220,122 @@ func (dev *VirtioSerial) FromOptions(options []option) error {
...
+			width, err := strconv.Atoi(option.value)
+			if err != nil || width < 1 {
+				return fmt.Errorf(fmt.Sprintf("Invalid value for virtio-gpu %s: %s", option.key, option.value))
+			}
+
+			dev.Width = width
+		}
+	}
+
+	if dev.Width < 1 && dev.Height < 1 {
```
cfergeau, line 331: any reason for checking for `< 1` instead of `== 0`?

cmd/vfkit/main.go
```diff
@@ -168,18 +184,35 @@ func runVirtualMachine(vmConfig *config.VirtualMachine, vm *vz.VirtualMachine) e
...
+			// errVMStateTimeout -> keep looping
 		}
-		if !errors.Is(err, errVMStateTimeout) {
-			log.Infof("virtualization error: %v", err)
-			return err
+	}()
+
+	for _, gpuDev := range vmConfig.VirtioGPUDevices() {
+		if gpuDev.UsesGUI {
+			runtime.LockOSThread()
```
cfergeau, line 207: If I'm not mistaken, `runtime.LockOSThread()` was already called at the beginning of the current function, so this is not really required? I agree it does not hurt to have it here as it's the place that really requires it.

EDIT: without the `runtime.LockOSThread()` call at the beginning of this function, vfkit crashes, so it's not only `StartGraphicApplication` which requires this.

cmd/vfkit/main.go
```diff
@@ -59,6 +60,9 @@ func newBootloaderConfiguration(opts *cmdline.Options) (config.Bootloader, error
 }
 
 func newVMConfiguration(opts *cmdline.Options) (*config.VirtualMachine, error) {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
```
cfergeau, line 64: This did not seem required in my testing (latest macOS 13 on a m1 macbook), do you remember why you added this?

cfergeau (commented):
A bunch of additional comments now that I could look at the virtio-gpu code and test it, but overall looks good, thanks again!

cmd/vfkit/main.go
```diff
@@ -113,10 +117,22 @@ func waitForVMState(vm *vz.VirtualMachine, state vz.VirtualMachineState) error {
...
 	vzVMConfig, err := vf.ToVzVirtualMachineConfig(vmConfig)
 	if err != nil {
 		return err
 	}
+
+	if opts.UseGUI {
+		for _, gpuDev := range vmConfig.VirtioGPUDevices() {
+			if gpuDev.Device == config.VirtioGPUDisplayDevice {
+				gpuDev.UsesGUI = true
```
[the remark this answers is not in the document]
jakecorrenti: I didn't make any changes to the VirtioGPU type. The `UsesGUI` attribute is still there for each Virtio-gpu device, and is automatically set to false. It can only be set to true if the user provides the `--gui` flag which is initially checked here.

cmd/vfkit/main.go
jakecorrenti: My intent behind it was to leave it open to the possibility that additional graphics devices may be supported. But I can remove it and make the assumption that the only graphics device that will be added to the VM will be the display

pkg/config/virtio.go
jakecorrenti: Fixed.

pkg/config/virtio.go
jakecorrenti: Fixed.

cmd/vfkit/main.go
jakecorrenti: This isn't required, I must have left it there accidentally

cfergeau (commented):
Apart from this `display` comment, should be good to go in, thanks for the changes you've made so far!

cfergeau (approved):

PR #46, opened 2023-06-23
Add json serialization/deserialization

cfergeau:
This allows other projects to write a configuration file based on virtual
machine.

This is based on https://github.com/crc-org/vfkit/pull/43
These 2 commits could be merged in the initial one
```
   json: Fix segfault when json does not have a 'kind' field
   json: Fix 'MashallJSON()' typos
```

This improves on #43 by adding some unit tests, fixing a few bugs found by the
unit tests, and tightening error handling during unmarshalling.
I've also simplified the code/refactored it a bit.


cfergeau:
More `json` annotations to the struct used by vfkit could be useful, as now  the json is not always looking great with respect to capitalization.

baude (approved):

PR #49, opened 2023-06-28
vf: Add toVz() methods for all virtio devices

cfergeau:
This follows the same model as for virtio-fs or virtio-blk devices so that
it's possible to add multiple devices of the same type to the VM.
The only exception is virtio-vsock which is documented by Apple as only
being allowed once.


baude:
LGTM

baude (approved):

PR #50, opened 2023-07-06
virtionet: Add gvproxy support

cfergeau:
This fixes --device virtio-net,unixSocketPath= which was non-functional 
until now, it was not using the correct socket type ("unix" instead of
"unixgram"). This commit fixes this. Immediately after
DialUnix("unixgram"), vfkit sends the magic string
"VFKT" so that the other end knows where to send the data to (there is no
`Accept` with unixgram sockets). There could be a better way but I'm not
aware of it.

### DIFF
--- a/pkg/vf/virtio.go
+++ b/pkg/vf/virtio.go
@@ -2,7 +2,6 @@ package vf
 
 import (
 	"fmt"
-	"net"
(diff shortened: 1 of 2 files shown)


baude (approved):
LGTM

PR #51, opened 2023-07-07
No CGO for rest consumers/clients

baude:
Refactored portions of the rest code to not require CGO for rest consumers.  Most of the refactored code was moved into pkg/rest/define to isolate it.

Also pruned an unused function and related unit test.

### DIFF
--- a/pkg/rest/config.go
+++ b/pkg/rest/config.go
@@ -1,10 +1,1 @@
 package rest
-
-type ServiceScheme int
-
-const (
-	TCP ServiceScheme = iota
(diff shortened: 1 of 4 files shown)


gbraad (approved):

cfergeau:
An alternative would be to use interfaces to separate the code-hex/vz code from the rest code, I tried this in https://github.com/cfergeau/vfkit/commit/c39510306f8ead5268b8e8329d16dc201b249e76

baude:
closing in favor of #52

PR #58, opened 2023-08-17
restore proper definition of width/height

ErwanMAS:
Width/Height for GPU parameter was switched.

### DIFF
--- a/cmd/vfkit/main.go
+++ b/cmd/vfkit/main.go
@@ -200,7 +200,7 @@ func runVirtualMachine(vmConfig *config.VirtualMachine, vm *vz.VirtualMachine) e
 	for _, gpuDev := range vmConfig.VirtioGPUDevices() {
 		if gpuDev.UsesGUI {
 			runtime.LockOSThread()
-			err := vm.StartGraphicApplication(float64(gpuDev.Height), float64(gpuDev.Width))
+			err := vm.StartGraphicApplication(float64(gpuDev.Width), float64(gpuDev.Height))
 			runtime.UnlockOSThread()
 			if err != nil {
 				return err
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -221,11 +221,11 @@ The share can be mounted in the guest with `mount -t virtiofs vfkitTag /mnt`, wi
 The `--device virtio-gpu` option allows the user to add graphical devices to the virtual machine.
 
 #### Arguments
-- `height`: the vertical resolution of the graphical device's resolution. Defaults to 800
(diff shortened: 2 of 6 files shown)


cmd/vfkit/main.go
```diff
@@ -200,7 +200,7 @@ func runVirtualMachine(vmConfig *config.VirtualMachine, vm *vz.VirtualMachine) e
 	for _, gpuDev := range vmConfig.VirtioGPUDevices() {
 		if gpuDev.UsesGUI {
 			runtime.LockOSThread()
-			err := vm.StartGraphicApplication(float64(gpuDev.Height), float64(gpuDev.Width))
+			err := vm.StartGraphicApplication(float64(gpuDev.Width), float64(gpuDev.Height))
```
cfergeau, line 203: Oops, good catch!

doc/usage.md
```diff
@@ -221,11 +221,11 @@ The share can be mounted in the guest with `mount -t virtiofs vfkitTag /mnt`, wi
 The `--device virtio-gpu` option allows the user to add graphical devices to the virtual machine.
 
 #### Arguments
-- `height`: the vertical resolution of the graphical device's resolution. Defaults to 800
-- `width`: the horizontal resolution of the graphical device's resolution. Defaults to 600
+- `height`: the vertical resolution of the graphical device's resolution. Defaults to 600
+- `width`: the horizontal resolution of the graphical device's resolution. Defaults to 800
```
cfergeau, line 225: Could you switch the order of height and width in most places so that width comes before height? Seeing 600 then 800 to mean 800x600 will confuse me every time ^^

doc/usage.md
```diff
@@ -221,11 +221,11 @@ The share can be mounted in the guest with `mount -t virtiofs vfkitTag /mnt`, wi
...
 
 #### Example
-`--device virtio-gpu,height=1920,width=1080`
+`--device virtio-gpu,height=1080,width=1920`
```
cfergeau, line 228: put width before height in these examples too

pkg/config/virtio_test.go
```diff
@@ -158,24 +158,24 @@ var virtioDevTests = map[string]virtioDevTest{
 		newDev: VirtioGPUNew,
 		expectedDev: &VirtioGPU{
 			false,
-			VirtioGPUResolution{800, 600},
+			VirtioGPUResolution{600, 800},
```
cfergeau, line 161: and I'd switch the fields order in VirtioGPUResolution too

cfergeau (commented):
Looks good to me, though it would be easier to read with width and then height in most places.

PR #60, opened 2023-08-28
Improve "is kernel compressed" check

cfergeau:
ARM64 kernels must be uncompressed before they can be booted.
vfkit has a check for this to warn the user before attempting to start the VM.
However this check has both false negatives and false positives (with an
uncompressed puipui linux kernel, and with a compressed fedora kernel)

This PR should fix both situations


pkg/vf/bootloader.go
```diff
@@ -22,17 +42,7 @@ func isKernelUncompressed(filename string) (bool, error) {
...
-		return false, err
-	}
-	// uncompressed ARM64 kernels are matched as a MS executable, which is
-	// also an archive, so we need to special case it
-	if kind == matchers.TypeExe {
-		return true, nil
-	}
-
-	return false, nil
+	return isUncompressedArm64Kernel(buf), nil
```
praveenkumar, line 45: Where this `buf` variable comes from?

praveenkumar: nvm, got it.

praveenkumar:
CI is failing as part of linter.

cfergeau:
> CI is failing as part of linter.

Fixed now!

PR #61, opened 2023-09-12
tests: Add basic e2e test

cfergeau:
This reuses the puipui linux work from Code-Hex/vz.
The OS is abstracted with a OSProvider interface which will hopefully
be implementable with Fedora CoreOS or other distros.

Tests which are currently implemented:
- start a VM, and make sure it's possible to connect through ssh (both
  over a regular network interface, and over vsock)
- for most devices, run lspci to make sure it's present in the VM
- vsock-listen and vsock-connect
- file sharing


test/vm_test.go
```diff
@@ -0,0 +1,260 @@
...
+		},
+	},
+	"virtio-fs": {
+		vendorID: 0x1af4, // Red Hat
+		deviceID: 0x105a,
+		createDev: func(t *testing.T) (config.VirtioDevice, error) {
+			return config.VirtioFsNew("./", "vfkit-share-test")
+		},
+	},
+	"virtio-gpu": {
```
cfergeau, line 205: `virtio-gpu`, `virtio-input/*` should only be run on macOS 13+ as these devices are not supported with earlier virtualization framework versions.

Makefile
```diff
@@ -9,8 +9,9 @@ include tools/tools.mk
 
 build: out/vfkit
 
-test:
-	go test ./pkg/...
+test: build
+	@go test ./pkg/...
+	@go test -timeout 20m ./test
```
anjannath, line 14: `go test -v` output is a bit better, it shows which tests are running for each package

anjannath:
one thing i noticed is that the tests are downloading the puipui linux tarball multiple times (one download for each test scenario) which is fine since it is very small in size, but if we do test fedora or some other real distro we need to implement OSProvider to download the image only once

anjannath (approved):

cfergeau:
/hold
added a commit to try to understand why gh actions don't behave as I expect them to.

cfergeau:
/unhold
hopefully should be good this time. Tests can run/succeed on macos 12/13, and are skipped on macos 11

cfergeau:
> one thing i noticed is that the tests are downloading the puipui linux tarball multiple times (one download for each test scenario) which is fine since it is very small in size, but if we do test fedora or some other real distro we need to implement OSProvider to download the image only once

Yep, this definitely needs improvements :)

Makefile
cfergeau: I'm not a big fan of the -v option, as it tends to be overly verbose, and test failures end up buried in the verbose output, instead of nicely standing out when `-v` is not used.
I've changed the PR to add this regardless.

PR #63, opened 2023-10-25
feat(rest): add can* params in `GET /vm/state`

BlackHole1:
In the previous API, there was no way to detect before changing the status, which would make it difficult for developers to handle the following situations:

1. When the virtual machine status is Stop, calling Pause may fail.
2. When the virtual machine status is Pause, setting the status to Pause again may fail.

In the previous solution, developers needed to first get the current status of the virtual machine and then decide whether to proceed to the next step. However, the current virtual machine status is very diverse (see: https://github.com/Code-Hex/vz/blob/bd29a7ea3d39465c4224bfb01e990e8c220a8449/virtualization.go#L23), which would require developers to handle various cases perfectly. This PR solves this problem.

### DIFF
--- a/pkg/rest/vf/state_change.go
+++ b/pkg/rest/vf/state_change.go
@@ -53,3 +53,23 @@ func (vm *VzVirtualMachine) HardStop() error {
 	logrus.Debug("force stopping machine")
 	return vm.VzVM.Stop()
 }
+
+func (vm *VzVirtualMachine) CanStart() bool {
+	return vm.VzVM.CanStart()
+}
+
+func (vm *VzVirtualMachine) CanPause() bool {
+	return vm.VzVM.CanPause()
+}
+
+func (vm *VzVirtualMachine) CanResume() bool {
+	return vm.VzVM.CanResume()
+}
+
+func (vm *VzVirtualMachine) CanStop() bool {
+	return vm.VzVM.CanRequestStop()
+}
+
+func (vm *VzVirtualMachine) CanHardStop() bool {
+	return vm.VzVM.CanStop()
+}
--- a/pkg/rest/vf/vm_config.go
+++ b/pkg/rest/vf/vm_config.go
@@ -33,7 +33,14 @@ func (vm *VzVirtualMachine) Inspect(c *gin.Context) {
 // getVMState retrieves the current vm state
 func (vm *VzVirtualMachine) GetVMState(c *gin.Context) {
 	current := vm.GetState()
-	c.JSON(http.StatusOK, gin.H{"state": current.String()})
+	c.JSON(http.StatusOK, gin.H{
+		"state":       current.String(),
+		"canStart":    vm.CanStart(),
+		"canPause":    vm.CanPause(),
+		"canResume":   vm.CanResume(),
+		"canStop":     vm.CanStop(),
+		"canHardStop": vm.CanHardStop(),
+	})
 }
 
 // setVMState requests a state change on a virtual machine.  At this time only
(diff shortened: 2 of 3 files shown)


cfergeau:
> developers needed to first get the current status of the virtual machine and then decide whether to proceed to the next step.

One minor comment, such an API will be racy, the VM state can change between the `canXXX` query and the `change state` query.

I'm not sure `/vm/can/:state` is the best API for this, maybe this could be added to `/vm/state`?
```
{"state": "string", "canStop": "true", "canPause": "false", "canForceStop": "true", "canResume": "..."}
```

Or when changing state, we could use different errors for "the operation is impossible in this state" and "there was an error when performing the operation"?

BlackHole1:
> One minor comment, such an API will be racy, the VM state can change between the canXXX query and the change state query.

Good catch! I have changed the code.

pkg/rest/vf/state_change.go
```diff
@@ -54,40 +54,18 @@ func (vm *VzVirtualMachine) HardStop() error {
...
 
 func (vm *VzVirtualMachine) CanStop() bool {
-	can := vm.VzVM.CanRequestStop()
-	logrus.Debug("can stop: ", can)
-	return can
+	return vm.VzVM.CanRequestStop()
 }
 
 func (vm *VzVirtualMachine) CanHardStop() bool {
-	logrus.Debug("can hard top: ", true)
```
cfergeau: there's a `vm.vzVM.CanStop()`

pkg/rest/vf/vm_config.go
```diff
@@ -34,7 +33,13 @@ func (vm *VzVirtualMachine) Inspect(c *gin.Context) {
 // getVMState retrieves the current vm state
 func (vm *VzVirtualMachine) GetVMState(c *gin.Context) {
 	current := vm.GetState()
-	c.JSON(http.StatusOK, gin.H{"state": current.String()})
+	c.JSON(http.StatusOK, gin.H{
+		"state":       current.String(),
+		"canPause":    vm.CanPause(),
+		"canResume":   vm.CanResume(),
+		"canStop":     vm.CanStop(),
+		"canHardStop": vm.CanHardStop(),
```
cfergeau, line 41: for completeness, I'd add `vm.CanStart()`

pkg/rest/vf/state_change.go
BlackHole1: In fact, our `Stop` corresponds to `RequestStop` in vz, and `HardStop` corresponds to `Stop` in vz. In vz, there is no API called `CanStop` (because powering off is always possible at any time).

BlackHole1: Good catch. I used to think that vz did not provide this API.😅

BlackHole1:
@cfergeau Done

cfergeau:
> @cfergeau Done

I've squashed all commits to a single one, and copied the PR description to the commit, and force-pushed to your `feat-can` branch, I hope that's fine with you!

cfergeau:
/lgtm

(I'll approve after CI runs)

BlackHole1:
> > @cfergeau Done
> 
> I've squashed all commits to a single one, and copied the PR description to the commit, and force-pushed to your `feat-can` branch, I hope that's fine with you!

NP. In the future, I will pay attention to this when I submit a PR.

cfergeau:
> NP. In the future, I will pay attention to this when I submit a PR.

The expectations vary between projects, I don't mind doing the adjustments. I did it myself to avoid additional back & forth

cfergeau (approved):

PR #65, opened 2023-10-27
Set vfkitVersion to 0.5.0

cfergeau:


cmd/vfkit/root.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const vfkitVersion = "0.1.1"
+const vfkitVersion = "0.5.0"
```
anjannath, line 12: we skipped some numbers in between, is `0.5.0` used to mean a development release?

anjannath (approved):

cmd/vfkit/root.go
cfergeau: I added an explanation to the commit log, but when I made the 0.1.1 release, there were some comments that 0.1.1 feels like a very early version, kind of 'alpha' software.

cfergeau: The aim is to have 1.0 sooner rather than later :)

anjannath: got it, i thought maybe you used odd/even versioning :)

PR #68, opened 2023-11-22
Fix crash in timesync code

cfergeau:
When ConnectVSockSync fails in watchWakeupNotifications(), the returned
net.Conn contains a nil value, but is not a nil interface
(ie checking it for `== nil` returns false). This means we'll try to call
`Close()` on the connection, which will cause a nil pointer dereference. 
See https://go.dev/doc/faq#nil_error and
https://groups.google.com/g/golang-nuts/c/wnH302gBa4I for details about
this behaviour.

This bug is fixed by returning a nil interface when
`vsockDevice.Connect(uint32(port))` returns an error.


pkg/vf/vsock.go
```diff
@@ -25,7 +25,13 @@ func ConnectVsockSync(vm *vz.VirtualMachine, port uint) (net.Conn, error) {
...
 	vsockDevice := socketDevices[0]
 
-	return vsockDevice.Connect(uint32(port))
+	conn, err := vsockDevice.Connect(uint32(port))
+	if err != nil {
+		// we can't `return vsockDevice.Connect()` directly, see https://go.dev/doc/faq#nil_error
+		// checking the return value for nil won't work as expected if we don't do this
+		return nil, err
+	}
+	return conn, nil
```
praveenkumar, line 34: 👍🏼

praveenkumar (approved):

PR #69, opened 2023-11-23
Generate vfkit version from git tag

cfergeau:
Currently the vfkit version must be manually set
before making a release, which is error-prone (this was the reason for the
0.1.1 release).
This PR generates it from git tags instead using git describe.
It also uses some git-archive magic to set the proper version
when a tarball is used for building vfkit.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/pkg/cmdline/version.go export-subst
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -21,6 +21,15 @@ jobs:
     steps:
       - name: Check out repository code
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: >-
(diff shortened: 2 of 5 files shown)


anjannath (commented):
This is missing the `.gitattributes` file

cfergeau:
I added the missing file. This also needs slightly more work when  building in github actions, see https://github.com/containers/gvisor-tap-vsock/pull/314/commits/cf2df8904fa55a771609319780a8f404b812d54a

anjannath (approved):

cfergeau:
> > I added the missing file. This also needs slightly more work when building in github actions, see [containers/gvisor-tap-vsock@cf2df89](https://github.com/containers/gvisor-tap-vsock/commit/cf2df8904fa55a771609319780a8f404b812d54a)
> 
> so on github actions job we have to manually fetch the tags, thing to remember for when we add a automatic release job to this repo

I've now adjusted the gh actions workflows to have proper release information on tag pushes and on git builds.

PR #72, opened 2023-12-26
refactor(error): remove unnecessary error return

BlackHole1:
If a function does not produce any errors, it should not return an error (except for functions defined by an interface). This will reduce the mental burden on users.

The AddDevices method was also added, on the side.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -142,10 +142,8 @@ func (vm *VirtualMachine) VirtioVsockDevices() []*VirtioVsock {
 
 // AddDevice adds a dev to vm. This device can be created with one of the
 // VirtioXXXNew methods.
-func (vm *VirtualMachine) AddDevice(dev VirtioDevice) error {
+func (vm *VirtualMachine) AddDevice(dev VirtioDevice) {
 	vm.Devices = append(vm.Devices, dev)
-
-	return nil
 }
 
 func (vm *VirtualMachine) AddTimeSyncFromCmdLine(cmdlineOpts string) error {
@@ -165,10 +163,10 @@ func (vm *VirtualMachine) TimeSync() *TimeSync {
 	return vm.Timesync
 }
 
-func TimeSyncNew(vsockPort uint) (VMComponent, error) {
+func TimeSyncNew(vsockPort uint) VMComponent {
 	return &TimeSync{
 		VsockPort: vsockPort,
-	}, nil
+	}
 }
 
 func (ts *TimeSync) ToCmdLine() ([]string, error) {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -143,7 +143,12 @@ func (vm *VirtualMachine) VirtioVsockDevices() []*VirtioVsock {
 // AddDevice adds a dev to vm. This device can be created with one of the
 // VirtioXXXNew methods.
 func (vm *VirtualMachine) AddDevice(dev VirtioDevice) {
-	vm.Devices = append(vm.Devices, dev)
+	vm.AddDevices(dev)
+}
+
+// AddDevices adds a list of devices to vm.
+func (vm *VirtualMachine) AddDevices(dev ...VirtioDevice) {
+	vm.Devices = append(vm.Devices, dev...)
 }
 
 func (vm *VirtualMachine) AddTimeSyncFromCmdLine(cmdlineOpts string) error {
--- a/pkg/config/json_test.go
+++ b/pkg/config/json_test.go
@@ -79,14 +79,12 @@ var jsonTests = map[string]jsonTest{
 			dev = VirtioVsockNew(1234, "/virtiovsock", false)
 			vm.AddDevice(dev)
 			// virtio-fs
-			dev = VirtioFsNew("/virtiofs", "tag")
-			vm.AddDevice(dev)
+			fs := VirtioFsNew("/virtiofs", "tag")
 			// USB mass storage
-			dev = USBMassStorageNew("/usbmassstorage")
-			vm.AddDevice(dev)
+			usb := USBMassStorageNew("/usbmassstorage")
 			// rosetta
-			dev = RosettaShareNew("vz-rosetta")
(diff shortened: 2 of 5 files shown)


gbraad:
Thank you. I'll let @cfergeau decide as there might have been reasons, though he is currently on leave.

gbraad:
> The AddDevices method was also added, on the side. (If you are concerned about this, I will move it to another PR.)

Knowing Christophe, this should have been at least a separate commit, but best to make a PR specific to this with clear description.

gbraad:
Waiting for repush to ok-for-test

BlackHole1:
> this should have been at least a separate commit

Done

pkg/config/json_test.go
```diff
@@ -62,55 +55,38 @@ var jsonTests = map[string]jsonTest{
...
+			vm.AddDevice(dev)
 			// virtio-gpu
-			dev, err = VirtioGPUNew()
-			require.NoError(t, err)
-			err = vm.AddDevice(dev)
+			dev = VirtioGPUNew()
 			require.NoError(t, err)
+			vm.AddDevice(dev)
 			// virtio-net
 			dev, err = VirtioNetNew("00:11:22:33:44:55")
```
cfergeau, line 69: This is the main reason why most `*New()` methods return an error, some of them require an error return, so for me it's better if they are consistent and all return an error, instead of having to remember it. This also makes the API future-proof if one day we need to return an error.

pkg/config/config.go
```diff
@@ -142,10 +142,8 @@ func (vm *VirtualMachine) VirtioVsockDevices() []*VirtioVsock {
 
 // AddDevice adds a dev to vm. This device can be created with one of the
 // VirtioXXXNew methods.
-func (vm *VirtualMachine) AddDevice(dev VirtioDevice) error {
+func (vm *VirtualMachine) AddDevice(dev VirtioDevice) {
 	vm.Devices = append(vm.Devices, dev)
```
cfergeau, line 146: Yep, this one is simpler without the error.

cfergeau (commented):
Sorry for the late reviews, you filed these PRs just before I went away for a while :)
I'll need to think some more about the error removal from the `*New` changes, there are some plus and minuses to having it/not having it :)

pkg/config/config.go
```diff
@@ -146,6 +146,11 @@ func (vm *VirtualMachine) AddDevice(dev VirtioDevice) {
...
 }
 
+// AddDevices adds a list of devices to vm.
```
cfergeau, line 149: This is what `AddDevice` should have been ;) maybe change it to `return AddDevices(dev)`

BlackHole1: Good idea!

BlackHole1:
Hey @cfergeau. I agree with what you said: maintaining a consistent style is also a way to reduce cognitive load.

In this current PR, I noticed that there are only two `FooNew()` methods that return an error, and I’m wondering if we could also make them not return an error to maintain a consistent style.

1. `VirtioInputNew`: This method calls `dev.validate()`, but it actually doesn’t need to because other `FooNew()` methods will call `dev.validate()` during the `ToCmdLine` and `FromOptions` stages. So it’s pretty simple to make this method *no-err*.
2. `VirtioNetNew`: This method needs to call `net.ParseMAC` to determine whether the input mac address is correct. If we change the function signature to: `VirtioNetNew(macAddress net.HardwareAddr)`, could we avoid having this function return an error?

-----

as you said: "This also makes the API future-proof if one day we need to return an error". is also a reason. But for me: The optimal design for now will surely not be optimal in the future, avoid over-design. When the future returns an error, we can consider ways to improve, such as changing the function name to: `FooSetup`. At this stage, `FooNew` only does one thing, which is to create a struct instance.

pkg/rest/vf/vm_config.go
```diff
@@ -18,7 +18,7 @@ func NewVzVirtualMachine(vm *vz.VirtualMachine, config *vz.VirtualMachineConfigu
 	return &VzVirtualMachine{config: config, VzVM: vm}
 }
 
-// inspect returns information about the virtual machine like hw resources
+// Inspect returns information about the virtual machine like hw resources
```
cfergeau, line 21: Can you split these comment changes in a separate commit?

cfergeau:
> as you said: "This also makes the API future-proof if one day we need to return an error". is also a reason. But for me: The optimal design for now will surely not be optimal in the future, avoid over-design. When the future returns an error, we can consider ways to improve, such as changing the function name to: `FooSetup`. At this stage, `FooNew` only does one thing, which is to create a struct instance.

For internal/private API, I agree. For external API, changing the API in incompatible ways has to be balanced with the annoyances this will cause to users.
Removing the errors mean changing https://github.com/crc-org/crc/blob/2b661eb9eb2c57676e1f18b728e8aa5c025ba360/pkg/drivers/vfkit/driver_darwin.go#L187-L196 , changing https://github.com/containers/podman/blob/01b2243e73076a0e71a667153af1b98ecbf3b2f5/pkg/machine/applehv/vfkit.go#L10-L33

And I don't know if there are other users of vfkit's config code.
If we make the change now, and have to change it back in the future to add the error, then we'll have broken our API twice, and annoyed twice vfkit go users. This is why I'm very conservative with these changes. If we don't have a choice, I'll consider it, otherwise, I'll think hard about it before making the change.

(and API changes get even harder after the module has had a v1 release, as breaking API means we have to change the module path, which causes even more churn to go programs using vfkit)

BlackHole1:
@cfergeau Okay, you convinced me. I will close this PR and create two new ones to address the following issues:

1. Add the `AddDevices` method
2. Improve comments

cfergeau:
> @cfergeau Okay, you convinced me. I will close this PR and create two new ones to address the following issues:
> 
>     1. Add the `AddDevices` method
> 
>     2. Improve comments

Sharing our different point of views made for an interesting discussion! Thanks for the 2 additional PRs :)

PR #75, opened 2023-12-27
refactor: improve code for wait vm to reach the specified state

BlackHole1:
In previous implementations, there was a need to maintain a timeout error variable to ignore while waiting for the VM to stop.
This part of the code was not intuitive enough, and other maintainers couldn’t grasp its intention at first glance.
This improvement has optimized the `waitForVMState` function by adding a timeout parameter, making the code more intuitive.

### DIFF
--- a/cmd/vfkit/main.go
+++ b/cmd/vfkit/main.go
@@ -20,7 +20,6 @@ limitations under the License.
 package main
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"os/signal"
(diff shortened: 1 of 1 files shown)


cfergeau (approved):
The main reason for this timeout at start time was to avoid an infinite loop if there's a virtualization error early.
But after https://github.com/crc-org/vfkit/commit/52a4bc5d76fb976d9b66106e349b98ef1bfbb0b9 I'm not sure this timeout is still useful, it can probably be removed.

These changes look good to me as is, I'll need to test them before merging.

gbraad:
@cfergeau take this after this week's meeting, or let us test this as part of the other changes?

PR #76, opened 2024-01-06
virtio: Enable full disk caching

cgwalters:
We're seeing highly reliable disk corruption in podman machine with the default configuration, and this
fixes it for me.

This looks like the same thing as
https://github.com/lima-vm/lima/commit/488c95c41c67ae0c3afa8f35173a6e1ef59d29ef

### DIFF
--- a/pkg/vf/virtio.go
+++ b/pkg/vf/virtio.go
@@ -275,7 +275,9 @@ func (config *StorageConfig) toVz() (vz.StorageDeviceAttachment, error) {
 	if config.ImagePath == "" {
 		return nil, fmt.Errorf("missing mandatory 'path' option for %s device", config.DevName)
 	}
-	return vz.NewDiskImageStorageDeviceAttachment(config.ImagePath, config.ReadOnly)
+	syncMode := vz.DiskImageSynchronizationModeFsync
+	caching := vz.DiskImageCachingModeCached
+	return vz.NewDiskImageStorageDeviceAttachmentWithCacheAndSync(config.ImagePath, config.ReadOnly, caching, syncMode)
 }
 
 func (dev *USBMassStorage) toVz() (vz.StorageDeviceConfiguration, error) {


pkg/vf/virtio.go
```diff
@@ -275,7 +275,9 @@ func (config *StorageConfig) toVz() (vz.StorageDeviceAttachment, error) {
 	if config.ImagePath == "" {
 		return nil, fmt.Errorf("missing mandatory 'path' option for %s device", config.DevName)
 	}
-	return vz.NewDiskImageStorageDeviceAttachment(config.ImagePath, config.ReadOnly)
+	syncMode := vz.DiskImageSynchronizationModeFsync
```
cgwalters, line 278: In the future it'd make sense I guess to plumb through these options as API too, but given the current foot-gun nature of things, let's hardcode for now.

gbraad:
Thanks. We also spoke with Sergio Lopez and got confirmed that this is most likely caused by caching/not flushing in time. We had several reports that this works for them.

gbraad (approved):

cfergeau:
Was wondering about performance impact, but tests in https://github.com/lima-vm/lima/pull/2026#issuecomment-1834136049  say there is no impact.

cfergeau:
https://github.com/lima-vm/lima/issues/1957
https://github.com/utmapp/UTM/issues/4840
contain a lot of useful information.

jorhett:
This is a very important fix. When can we expect a release?

cfergeau:
> This is a very important fix. When can we expect a release?

I'm aiming to cut a release this week. In the mean time, I've already added the patch to this brew recipe https://github.com/cfergeau/homebrew-crc/blob/main/vfkit.rb

Are you also hitting this bug?

jorhett:
> Are you also hitting this bug?

No, I just recently discovered from a comment on another bug that applehv support was now available in Podman. I was poking around to see why it wasn't announce or visible in the docs (other than a mention of it as a valid provider) and stumbled on this bug.  Figured I should wait for this to be in the release before I started recommending that our engineers give this a try.  If it's already in the brew recipe, that may suffice.

What's your feeling about stability versus Qemu?  Would you turn a few hundred engineers loose on this?

gbraad:
I wanted to add a comment in answer to:

>  just recently discovered from a comment on another bug that applehv support was now available in Podman

but saw Christophe already did. Some are still fixing some race conditions with applehv, as the implementations differs slightly with CRC. We hope to resolve this soon and converge; just note that this also needs a gvproxy release, which is more closely coordinated with Podman's release schedule.

PR #78, opened 2024-01-07
Add support for nvme devices

cgwalters:
This also seems to avoid the disk corruption that we see with virtio-blk; it reportedly has a small performance hit for raw speed, but I think avoiding the double caching (guest and host) is much better from a performance perspective.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -158,6 +158,23 @@ This adds a virtio-blk device to the VM which will be backed by the raw image at
 ```
 
 
+### NVM Express
+
+#### Description
+
+The `--device nvme` option adds a NVMe device to the virtual machine. The disk is backed by an image file on the host machine. This file is a raw image file.
+
+#### Arguments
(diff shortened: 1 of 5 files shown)


cgwalters:
Tested this e2e in concert with https://github.com/containers/podman/pull/21208 and I am also not seeing any corruption.

gbraad:
the NVME device is only supported on macOS 14, so this would error on 13.

cgwalters:
Your fixup LGTM

cfergeau:
/lgtm
/approve
/hold
(holding this to leave time for the tests to run)

cfergeau:
Interestingly, Apple recommends using virtio-blk https://developer.apple.com/videos/play/wwdc2023/10007/?time=630
They describe NVMe as a fallback for linux VM images which don't have the virtio-blk module built-in, and need an alternative way to access the disk data.

PR #79, opened 2024-01-08
config: Document that MemoryBytes is in MiB

cgwalters:
I lost about 10 minutes figuring this out.


cgwalters:
Hm wait...maybe I'm wrong and confused...

PR #89, opened 2024-01-25
feat(config): add AddDevices function

BlackHole1:
In most cases, there might be more than one device. Adding the AddDevices method helps reduce the amount of code.


cfergeau:
/ok-to-test
/lgtm
/approve
/hold

(holding so that the tests run before we merge it)

PR #103, opened 2024-02-22
json: test: Add 'stability' JSON test

cfergeau:
We don't want the JSON serialization of pkg/config/ data structures to
change between releases.
This commit makes use of the `reflect` package to test the JSON serialization
of as many fields as possible. If one of these field change, the test
will make us aware of it.
We still need to explicitly list the types to be tested (ie the types
which can be serialized to JSON).

### DIFF
--- a/pkg/config/json_test.go
+++ b/pkg/config/json_test.go
@@ -2,10 +2,64 @@ package config
 
 import (
 	"encoding/json"
+	"fmt"
+	"reflect"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 )
+
+func contains(strings []string, val string) bool {
+	for _, str := range strings {
+		if val == str {
+			return true
+		}
+	}
+
+	return false
+}
+
+// This sets all the fields of the `obj` struct to non-empty values.
+// This will be used to test JSON serialization as extensively as possible to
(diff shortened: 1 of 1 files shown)


BlackHole1 (approved):
Code LGTM.
The only concern I have is that if we add new devices in the future, we need to make sure to add corresponding test cases in `jsonStabilityTests`. There is a possibility that we might forget to do so.

cfergeau:
> The only concern I have is that if we add new devices in the future, we need to make sure to add corresponding test cases in `jsonStabilityTests`. There is a possibility that we might forget to do so.

Yes, hopefully this won't happen too often. I've tried to find a programmatic way of listing all the types defined in `pkg/config`, but this does not seem trivial unfortunately.  I suspect something is doable with https://pkg.go.dev/go/types and maybe https://pkg.go.dev/go/importer.

BlackHole1:
> I suspect something is doable with https://pkg.go.dev/go/types and maybe https://pkg.go.dev/go/importer.
It seems a bit complicated, we can consider putting it on hold for now and reevaluate if we encounter this issue in the future.

cfergeau:
/approve 
(adding approved label after Kevin's and Praveen's reviews)

Issue #48, opened 2023-06-28
Access guest console when it is running


praveenkumar:
As of now we https://github.com/crc-org/vfkit/blob/c9a4b0861530ce7320aae34ad2394f41e2fd5e83/pkg/vf/virtio.go#L272-L276 connects the VM serial output to the stdin/stdout of the terminal where vfkit is running but there is no way to connect to it so that we can access VM console even in case of no ssh connectivity. 

https://developer.apple.com/documentation/virtualization/vzfilehandleserialportattachment?language=objc => this api allows bidirectional communication using file handles but we don't know how to consume it. 

This is important for crc side to debug some of the issues when ssh connection is broken.

cfergeau:
hyperkit has such support, I think the code they use for this is https://github.com/moby/hyperkit/blob/3cb0d5475244761c61e6fd4a562702c26b46f846/src/lib/uart_emul.c#L738-L784

The code for https://pkg.go.dev/github.com/pkg/term/termios#Pty is very similar: https://github.com/pkg/term/blob/v1.1.0/termios/pty.go#L22

cfergeau:
Seems to be relatively easy to do https://github.com/cfergeau/vfkit/tree/pty
This is a hack which changes  `--device virtio-serial,stdio` to output to a pseudoterminal instead of stdio.
Need to add cmdline parsing/... to this proof of concept.

cdrage:
Testing with https://github.com/cfergeau/vfkit/tree/pty I made some awesome progress on integrating some of this within PD!

I did encounter some issues however:

* Running the same command twice on a different image, I would get a different ttys which is to be expected.. but I couldn't access it / read / write.

I was using this command:

```sh
./vfkit-arm64 --cpus 2 --memory 2048 \
    --bootloader efi,variable-store=./efi-variable-store,create \
    --device virtio-blk,path=/Users/cdrage/bootc/image/disk.raw \
    --device virtio-serial,pty \
    --device virtio-net,nat,mac=72:20:43:d4:38:62 \
    --device virtio-rng \
    --device virtio-input,keyboard \
    --device virtio-input,pointing
 ```
 
 * It would be nice to specify exactly which path the ttys will go (be able to pick /dev/ttys123, etc.)
 
 * Unable to see boot screen / GRUB, only was able to see the login prompt.

**Comment by cfergeau** [comment]
You can remove
```
    --device virtio-input,keyboard \
    --device virtio-input,pointing
```
[...]

PR #113, opened 2024-03-15
serial: Add --device virtio-serial,pty support

cfergeau:
This new virtio-serial option allocates a pseudo-tty for the VM console. It
can then be accessed using `screen` for example. This is a bit similar to
the `--device virtio-serial,stdio` option, except that the console is not
tied to the terminal running vfkit, it's possible to connect/disconnect
from the pseudo-tty from any terminal.

This fixes https://github.com/crc-org/vfkit/issues/48


anjannath:
tested and working as expected..

```
% ./out/vfkit \
    --cpus 2 --memory 2048 \
    --bootloader efi,variable-store=efi-variable-store,create \
    --device virtio-blk,path=vfkit-test-image.raw --device virtio-serial,pty
INFO[0000] &{2 2048    {[efi variable-store=efi-variable-store create] true}  [virtio-blk,path=vfkit-test-image.raw virtio-serial,pty] none://  false}
INFO[0000] boot parameters: &{EFIVariableStorePath:efi-variable-store CreateVariableStore:true}
INFO[0000]
INFO[0000] virtual machine parameters:
INFO[0000] 	vCPUs: 2
INFO[0000] 	memory: 2048 MiB
INFO[0000]
INFO[0000] Adding virtio-blk device (imagePath: vfkit-test-image.raw)
INFO[0000] Using PTY (pty path: /dev/ttys002)
INFO[0000] virtual machine is running
INFO[0000] waiting for VM to stop
```

after this i was able to see the fedora login prompt by running `% screen /dev/ttys002`

praveenkumar:
I am testing it with crc bundle with the script part of contrib/script with following change but not able to get any login prompt
```
$ git diff
diff --git a/contrib/scripts/start-crc-bundle.sh b/contrib/scripts/start-crc-bundle.sh
old mode 100644
new mode 100755
index 34a28ee..965a497
--- a/contrib/scripts/start-crc-bundle.sh
+++ b/contrib/scripts/start-crc-bundle.sh
@@ -15,6 +15,6 @@ cp -c ${BUNDLE_PATH}/${DISKIMG} overlay.img
     --initrd "${BUNDLE_PATH}/${INITRD}" \
     --kernel-cmdline "${CMDLINE}" \
     --device virtio-blk,path=overlay.img \
-    --device virtio-serial,logFilePath=start-bundle.log \
+    --device virtio-serial,pty \ 
     --device virtio-net,nat,mac=72:20:43:d4:38:62 \
     --device virtio-rng
```


```
$ contrib/scripts/start-crc-bundle.sh /Users/prkumar/.crc/cache/crc_vfkit_4.13.14_arm64
[...]
INFO[0000]                                              
INFO[0000] virtual machine parameters:                  
INFO[0000] 	vCPUs: 2                                    
INFO[0000] 	memory: 2048 MiB                            
INFO[0000]                                              
INFO[0000] Adding virtio-blk device (imagePath: overlay.img) 
INFO[0000] Using PTY (pty path: /dev/ttys000)            <= tried to connect this with screen and nothing
INFO[0000] Adding virtio-net device (nat: true macAddress: [72:20:43:d4:38:62]) 
INFO[0000] Adding virtio-rng device                     
INFO[0000] virtual machine is running                   
INFO[0000] waiting for VM to stop  
```

I will try with fedora also same what @anjannath tested and put another comment.

praveenkumar:
To me even steps mentioned by @anjannath didn't work after `screen /dev/ttys000` I just see the blank without any prompt.

anjannath:
I tried again and it is working for me still, maybe you have to attach to the pty quickly otherwise you miss the login prompt and it doesn't re-appear?

https://github.com/crc-org/vfkit/assets/8885742/befe219b-cbc4-42ca-b8a5-3da6e50deafc

but another thing i noticed is that the keyboard only works sometimes, a few times (2 out of 4 runs) during testing keys are not echoed in screen, which i assumed to be the keyboard not working

praveenkumar:
> maybe you have to attach to the pty quickly otherwise you miss the login prompt and it doesn't re-appear?

I also tried to attach pty quickly but still nothing, waiting more that 5 mins.

praveenkumar:
> but another thing i noticed is that the keyboard only works sometimes, a few times (2 out of 4 runs) during testing keys are not echoed in screen, which i assumed to be the keyboard not working

So after 5-6 retry finally I am able to get the login prompt for fedora and crc bundle but then on crc I am not able to use the keyboard but on fedora able to do.

cfergeau:
> > but another thing i noticed is that the keyboard only works sometimes, a few times (2 out of 4 runs) during testing keys are not echoed in screen, which i assumed to be the keyboard not working
> 
> So after 5-6 retry finally I am able to get the login prompt for fedora and crc bundle but then on crc I am not able to use the keyboard but on fedora able to do.

Maybe [virtio: Simplify setRawMode](https://github.com/crc-org/vfkit/pull/113/commits/a570c1b0d0f4ae0e8f5c1d650d1466229d5fa13d) is causing issues. Or maybe it would be better to use https://github.com/pkg/term/blob/1a4a3b719465afccedae7788469c0b54a43417e0/termios/termios.go#L41-L50 , but I could not get this to work correctly.
Really not familiar with pty/terminal input/output :-/

praveenkumar:
> > > but another thing i noticed is that the keyboard only works sometimes, a few times (2 out of 4 runs) during testing keys are not echoed in screen, which i assumed to be the keyboard not working
> > 
> > 
> > So after 5-6 retry finally I am able to get the login prompt for fedora and crc bundle but then on crc I am not able to use the keyboard but on fedora able to do.
> 
> Maybe [virtio: Simplify setRawMode](https://github.com/crc-org/vfkit/pull/113/commits/a570c1b0d0f4ae0e8f5c1d650d1466229d5fa13d) is causing issues. Or maybe it would be better to use https://github.com/pkg/term/blob/1a4a3b719465afccedae7788469c0b54a43417e0/termios/termios.go#L41-L50 , but I could not get this to work correctly. Really not familiar with pty/terminal input/output :-/

I am also not sure what causing the issue but as of now this functionality is quite buggy and we might not able to consume it when required like debugging an broken ssh connection. Do you think we can use `tty` along with log to a file because in tty I also don't see boot logs ?

cfergeau:
The virtualization framework has an [`isConsole`](https://developer.apple.com/documentation/virtualization/vzvirtioconsoleportconfiguration/4024741-isconsole?language=objc) property which is not used in this PR, maybe it would help to set it.

cfergeau:
Superseded by #259 259

PR #141, opened 2024-05-09
Minor doc improvements.

williamtheaker:
This is a pretty minor change is mostly just capitalizing words like Linux.

The only change I think pointing out is in `doc/usage.md` where the truncate command is changed to use a flag that works on both the GNU implementation and the FreeBSD version shipped with macOS.

### DIFF
--- a/doc/missing-vz-api.md
+++ b/doc/missing-vz-api.md
@@ -1,13 +1,13 @@
-# Missing API
+# Missing APIs
 
-This document contains a non-exhaustive list of API which are not used/supported by `vfkit`, but which would be useful to have.
+This document contains a non-exhaustive list of APIs not currently used/supported by `vfkit`, but could be useful to have.
 
-## non-vz API
+## non-vz APIs
 
 - start vfkit process (integrating with https://pkg.go.dev/os/exec )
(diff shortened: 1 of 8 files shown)


doc/missing-vz-api.md
```diff
@@ -1,13 +1,13 @@
-# Missing API
+# Missing APIs
 
-This document contains a non-exhaustive list of API which are not used/supported by `vfkit`, but which would be useful to have.
+This document contains a non-exhaustive list of APIs not currently used/supported by `vfkit`, but could be useful to have.
```
cfergeau, line 3: "but could be useful to have", or "but which could be useful to have"? The latter makes more sense to me, but I'm not a native English speaker.

cfergeau (commented):
Thanks for the thorough cleanups! Minor wording question, apart from this this looks good to me!

doc/missing-vz-api.md
cfergeau: (and I suspect this doc is out of date by now, I don't think I've updated it after macOS 14 release)

williamtheaker:
```suggestion
This document is a non-exhaustive list of unimplemented APIs, which may be worth adding to `vfkit`.
```

How's this?

PR #156, opened 2024-06-14
Move linting to separate job.

williamtheaker:
Currently, each job in the matrix strategy runs golangci-lint, which is slightly inefficient and expensive given the price of macOS runners. This PR moves the lint step to a separate job and defaults to the stable version of Go.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -7,6 +7,16 @@ on:
       - "*"
   pull_request: {}
 jobs:
+  golangci:
+    name: lint
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: stable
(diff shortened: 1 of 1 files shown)


cfergeau (approved):

.github/workflows/compile.yml
```diff
@@ -7,6 +7,16 @@ on:
       - "*"
   pull_request: {}
 jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
```
cfergeau, line 12: This needs to run on macos, I think golangci-lint attempts to build some code, which can only be done on a macos runner for vfkit (see the CI failures)

williamtheaker, line 12:
```suggestion
    runs-on: macos-latest
```

williamtheaker: I switched this to `macos-latest`

PR #164, opened 2024-07-10
gh: Stop using macos11 runners

cfergeau:
This removes the macos11 usage from github actions, and starts using macos14
macos11 runners are deprecated or removed from github actions:
https://github.blog/changelog/2024-05-20-actions-upcoming-changes-to-github-hosted-macos-runners/

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -25,6 +25,7 @@ jobs:
         os:
         - macOS-12
         - macOS-13
+        - macOS-14
     steps:
       - name: Check out repository code
         uses: actions/checkout@v4
@@ -44,11 +45,12 @@ jobs:
       - name: Build
         run: make
       - name: Test
+        if: matrix.os != 'macOS-14'
         run: make test
       - name: vet
         run: go vet ./...
       - name: Upload vfkit artifact
-        if: matrix.os == 'macOS-13'
+        if: matrix.os == 'macOS-14'
         uses: actions/upload-artifact@v4
         with:
           name: Unsigned vfkit Universal Binary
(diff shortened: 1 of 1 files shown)


praveenkumar:
The test around macOS-14 is failing now.

praveenkumar:
> > The test around macOS-14 is failing now.
> 
> Yes it's being run on an arm64 node which can't do virt, so it's unfortunately an expected failure. I haven't found a way to force the run to be done on an amd64/macos14 node.

I don't think the runner for macos14 have intel one in the background only silicon available https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources which is on M1 (no virt support) . We either tweak the test case to run it on M1 or just don't add macOS-14 support for action.

cfergeau:
> > > The test around macOS-14 is failing now.
> > 
> > 
> > Yes it's being run on an arm64 node which can't do virt, so it's unfortunately an expected failure. I haven't found a way to force the run to be done on an amd64/macos14 node.
> 
> I don't think the runner for macos14 have intel one in the background only silicon available https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources which is on M1 (no virt support) . We either tweak the test case to run it on M1 or just don't add macOS-14 support for action.

I disabled the tests on macOS14.

praveenkumar:
Does this need rebase becuase lint is failing and using (1.60.2) ?

cfergeau:
> Does this need rebase becuase lint is failing and using (1.60.2) ?

lint is a mess at the moment, better to ignore it :-/

PR #179, opened 2024-08-18
Add readonly option for storage devices

nirs:
The virtio-blk, nvme, and usb-mass-storage devices can use now `readonly` option to create a read only device. This exposes the exiting StorageConfig.ReadOnly field.

Because this option is mostly useful to cdrom images, document the option only for usb-mass-storage.


cfergeau:
Thanks for the PR!

PR #185, opened 2024-08-27
Fix linting issues

cfergeau:
This PR updates golangci-lint to the latest version, and fixes the various new issues reported by new version, most notably the int overflow ones.
This also stops using the golangci-lint action to avoid automatic updates of golangci-lint, since these often need code changes which we might not be ready to make.
The changes will be done instead through a go.mod update. The PRs will be created automatically by dependabot, but we can choose when to merge the change.


tools/go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/crc-org/vfkit/tools
 
-go 1.21
+go 1.22.1
```
praveenkumar, line 3: @cfergeau once we take this in and then update `vfkit` for crc then on crc side also go 1.22.x will be required.

test/vm_helpers.go
```diff
@@ -139,7 +139,7 @@ type testVM struct {
 
 	sshNetwork     string
 	macAddress     string // for SSH over TCP
-	port           int
+	port           uint
```
praveenkumar, line 142: This file changes doesn't  look like part of this commit.

tools/go.mod
cfergeau: It's in `vfkit/tools`, not in `vfkit`, `crc` does not use `vfkit/tools` so this should not force a 1.22.x update in crc.
I tried with this in crc: `go mod edit --replace github.com/crc-org/vfkit=github.com/cfergeau/vfkit@lint` followed by `go mod tidy`, and go.mod is still using `go 1.21`.

test/vm_helpers.go
cfergeau: github UI does not make it easy to see which commit you were referring to :-/
There were some changes in the `gha:setup-go: Use go version from go.mod` which did not belong there, I moved these to the right commit (`lint: Fix potential vsock port overflows`).

praveenkumar:
looks like it need rebase 

other than that /lgtm

cfergeau:
> looks like it need rebase
> 
> other than that /lgtm

Rebased, which cleared the /lgtm /o\

praveenkumar:
Does this macOS-13 build failure is expected?

praveenkumar:
Also I just merged https://github.com/crc-org/vfkit/pull/164 one so let's rebase on top of it?

cfergeau:
> Does this macOS-13 build failure is expected?

It's not expected, let's see if re-running it will succeed (rebase does not seem required)

cfergeau:
> It's not expected, let's see if re-running it will succeed

All green this time!

PR #191, opened 2024-09-06
timesync: Fix automatic vsock device creation

cfergeau:
Timesync uses a vsock port for guest/host time synchronization.
When timesync is in use, if the VM has no vsock device, a new one is
automatically added.
However this got broken in commit a5db53d as we were generating the
hypervisor configuration before adding this device.

This means timesync only worked if you had another vsock device
configured.

### DIFF
--- a/pkg/vf/vm.go
+++ b/pkg/vf/vm.go
@@ -103,6 +103,17 @@ func (cfg *VirtualMachineConfiguration) toVz() (*vz.VirtualMachineConfiguration,
 			return nil, err
 		}
 	}
+	if cfg.config.Timesync != nil && cfg.config.Timesync.VsockPort != 0 {
+		// automatically add the vsock device we'll need for communication over VsockPort
+		vsockDev := VirtioVsock{
+			Port:   cfg.config.Timesync.VsockPort,
+			Listen: false,
+		}
+		if err := vsockDev.AddToVirtualMachineConfig(cfg); err != nil {
+			return nil, err
+		}
+	}
+
 	cfg.SetStorageDevicesVirtualMachineConfiguration(cfg.storageDevicesConfiguration)
 	cfg.SetDirectorySharingDevicesVirtualMachineConfiguration(cfg.directorySharingDevicesConfiguration)
 	cfg.SetPointingDevicesVirtualMachineConfiguration(cfg.pointingDevicesConfiguration)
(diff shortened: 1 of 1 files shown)


anjannath:
so the vsock config should be part of virtualMachineConfig before calling `cfg.SetSocketDevicesVirtualMachineConfiguration(cfg.socketDevicesConfiguration)`

anjannath (approved):

pkg/vf/vm.go
```diff
@@ -103,6 +103,17 @@ func (cfg *VirtualMachineConfiguration) toVz() (*vz.VirtualMachineConfiguration,
 			return nil, err
 		}
 	}
+	if cfg.config.Timesync != nil && cfg.config.Timesync.VsockPort != 0 {
+		// automatically add the vsock device we'll need for communication over VsockPort
+		vsockDev := VirtioVsock{
+			Port:   cfg.config.Timesync.VsockPort,
+			Listen: false,
+		}
```
anjannath, line 111: there's also a constructor `VirtioVsockNew`

anjannath: that constructor returns a different type `VMComponent` using that would would require us to cast the value to `VirtioVsock` again?

cfergeau: I tried to use it, but it's a bit messy, there are 2 types, `config.VirtioVsock` and `vf.VirtioVsock` which are aliases for the same type, if I use `config.VirtioVsockNew`, then I need to cast the result to `vf.VirtioVsock`. In addition to that, this returns an error I need to check, and this needs a cast from uint32 to uint.
Easier to manually create the `vf.VirtioVsock` we need even if I agree that it would have been nicer to be able to use `VirtioVsockNew` :)

cfergeau:
> so the vsock config should be part of virtualMachineConfig before calling `cfg.SetSocketDevicesVirtualMachineConfiguration(cfg.socketDevicesConfiguration)`

I've added this to the commit log to make it clearer.

pkg/vf/vm.go
cfergeau: > that constructor returns a different type `VMComponent` using that would would require us to cast the value to `VirtioVsock` again?

Only saw your question after adding my reply, but yes, we'd need an additional cast to the proper type

cfergeau:
> Since we are thinking about using [crc-org/crc#4341 (comment)](https://github.com/crc-org/crc/issues/4341#issuecomment-2331060283) way for timesync should we try that and drop it completely from this repo?

After we add `makestep ...` to the bundle and it works fine, we can consider dropping this timesync code. We are not there yet, so better to keep it working, especially as the fix is not complicated.

cfergeau:
> @cfergeau commit log has some file listings, that i guess got pasted while amending commit, could you please remove that

hrm yes, happens more often than not because of some broken process on my end. Fixed now.

anjannath (approved):

Issue #194, opened 2024-09-18
unix socket path length problems


cfergeau:
unix socket paths have a very limited length, 104 bytes on macos, 108 on linux

Podman is currently hitting this limit with vfkit:
```
  time="2024-08-20T10:16:30-05:00" level=info msg="Using unix socket /var/folders/l0/3hgwcpld7h3gzm5s82dm48_w0000gn/T/podman/30247524d8c0-gvproxy.sock"
  Error: dial unixgram /var/folders/l0/3hgwcpld7h3gzm5s82dm48_w0000gn/T/podman_test2150468280/Library/Application Support/vfkit/net-5639-2529860791.sock->/var/folders/l0/3hgwcpld7h3gzm5s82dm48_w0000gn/T/podman/30247524d8c0-gvproxy.sock: bind: invalid argument
```

This path is generated in https://github.com/crc-org/vfkit/blob/966a9282673dbe36c7fa1b0f2c09750839501d43/pkg/vf/virtionet.go#L21-L41

cfergeau:
We should also report a proper error when the path length is too long

cfergeau:
Possible solutions:
- allow to override the path the socket goes to with a local variable
- put the vfkit socket in the same directory as the socket it's connecting to

cfergeau:
Tentative patch for the 2nd option is
```
diff --git a/pkg/vf/virtionet.go b/pkg/vf/virtionet.go
index 72fd73c..28e257d 100644
--- a/pkg/vf/virtionet.go
+++ b/pkg/vf/virtionet.go
@@ -19,20 +19,12 @@ type VirtioNet struct {
 	localAddr *net.UnixAddr
 }
 
-func localUnixSocketPath() (string, error) {
-	homeDir, err := os.UserHomeDir()
+func localUnixSocketPath(dir string) (string, error) {
+	tmpFile, err := os.CreateTemp(dir, fmt.Sprintf("vfkit-%d-*.sock", os.Getpid()))
 	if err != nil {
 		return "", err
 	}
-	dir := filepath.Join(homeDir, "Library", "Application Support", "vfkit")
-	if err := os.MkdirAll(dir, 0755); err != nil {
-		return "", err
-	}
-	tmpFile, err := os.CreateTemp(dir, fmt.Sprintf("net-%d-*.sock", os.Getpid()))
-	if err != nil {
-		return "", err
-	}
-	// slightly racy, but this is in a directory only user-writable
[...]

PR #195, opened 2024-09-20
virtionet: Fix 'path too long' issues with virtio-net unixgram Support

cfergeau:
With unixgram sockets, we need to specify a client endpoint, which is a 
path on the filesystem with a maximum length of 104 bytes on macOS. When
using -device virtio-net,unixSocketPath=/tmp/vnet.sock, we need to create a
client endpoint before connecting to /tmp/vnet.sock. It's currently put
into `/Users/user/Library/Application Support`, which is already fairly
long. Depending on the username, the 104 bytes are easily exceeded, this
has been causing problems to podman machine.

This PR will move the client endpoint to the same dir as the server endpoint to give more control over its location to vfkit users. It will also report a better error when the endpoint path length is more than 104 bytes, and it tries to shorten the client endpoint filename.

This fixes https://github.com/crc-org/vfkit/issues/194


pkg/vf/virtionet.go
```diff
@@ -44,11 +48,14 @@ func (dev *VirtioNet) connectUnixPath() error {
...
 		Net:  "unixgram",
 	}
-	localSocketPath, err := localUnixSocketPath()
+	localSocketPath, err := localUnixSocketPath(filepath.Dir(dev.UnixSocketPath))
 	if err != nil {
 		return err
 	}
-	// FIXME: need to remove localSocketPath at process  exit
+	if len(localSocketPath) >= 104 {
+		return fmt.Errorf("socket path is too long: %d", len(localSocketPath))
```
baude, line 56: maybe add the max value here ?

cfergeau: I've changed it to
```
+       if len(localSocketPath) >= maxUnixgramPathLen {
+               return fmt.Errorf("unixgram path '%d' is too long: %d > %d bytes", localSocketPath, len(localSocketPath), maxUnixgramPathLen)
+       }
```

cfergeau:
I've finally updated the tests introduced in https://github.com/crc-org/vfkit/commit/d7b2799a7dc45cffc76bce5829be205efa2d7b0d

lstocchi (approved):
LGTM, tested and works fine

lstocchi:
The mac15 build fails though

lstocchi:
Maybe we can remove the macOS15 commit and add it in another PR?

lstocchi (approved):
LGTM

PR #197, opened 2024-09-30
doc: Set minimum supported version to macOS 12

cfergeau:
macOS 11 github runners are no longer available so we no longer build
and test vfkit on this platform.
macOS 11 has been out of support for a year now.
This commit changes the minimum supported version to macOS 12.
A mention of macOS 11 is kept, we'll remove it when we'll know for sure
vfkit is broken on this platform

This PR also adds some package-level go documentation to (hopefully) make it easier to get into the codebase


lstocchi (commented):
Everything looks correct to me 👍

PR #199, opened 2024-10-02
virtionet: testing connectUnixPath

lstocchi:
It adds tests for connectUnixPath and localUnixSocketPath func in virtionet.go.
It mainly test the use case where one end of the unixgram socket is longer of 104 bytes resulting in a failure.

### DIFF
--- a/pkg/vf/virtionet_test.go
+++ b/pkg/vf/virtionet_test.go
@@ -1,0 +1,84 @@
+package vf
+
+import (
+	"bytes"
+	"fmt"
+	"math/rand"
+	"net"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/crc-org/vfkit/pkg/config"
+	"github.com/stretchr/testify/require"
+)
+
+func testConnectUnixgram(t *testing.T) error {
+	unixSocketPath := filepath.Join("/tmp", fmt.Sprintf("vnet-test-%x.sock", rand.Int31n(0xffff))) //#nosec G404 -- no need for crypto/rand here
+	addr, err := net.ResolveUnixAddr("unixgram", unixSocketPath)
+	require.NoError(t, err)
+
+	l, err := net.ListenUnixgram("unixgram", addr)
+	require.NoError(t, err)
+
+	defer l.Close()
+	defer os.Remove(unixSocketPath)
+
+	dev := &VirtioNet{
+		&config.VirtioNet{
+			UnixSocketPath: unixSocketPath,
+		},
+		&net.UnixAddr{},
+	}
+
+	return dev.connectUnixPath()
+}
+
+func TestConnectUnixPath(t *testing.T) {
+	t.Run("Successful connection - no error", func(t *testing.T) {
+		err := testConnectUnixgram(t)
+		require.NoError(t, err)
+	})
+
+	t.Run("Failed connection - End socket longer than 104 bytes", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/vf/virtionet_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"strings"
+	"testing"
+
+	"github.com/crc-org/vfkit/pkg/config"
+)
+
+func testConnectUnixgram(t *testing.T) error {
+	unixSocketPath := filepath.Join("/tmp", fmt.Sprintf("vnet-test-%x.sock", rand.Int31n(0xffff))) //#nosec G404 -- no need for crypto/rand here
+	addr, err := net.ResolveUnixAddr("unixgram", unixSocketPath)
+	if err != nil {
```
cfergeau, line 18: In test code, we usuallly use `github.com/stretchr/testify` and `require.NoError(t, err)` which is more concise.

pkg/vf/virtionet_test.go
```diff
@@ -0,0 +1,100 @@
...
+		origUserHome := os.Getenv("HOME")
+		defer func() {
+			os.Setenv("HOME", origUserHome)
+		}()
+
+		// Create a string of 100 bytes to update the user home to be sure to create a socket path > 104 bytes
+		b := make([]byte, 100)
+		for i := range b {
+			b[i] = 'a'
+		}
```
cfergeau, line 58: This could use https://pkg.go.dev/bytes#Repeat

pkg/vf/virtionet_test.go
```diff
@@ -0,0 +1,100 @@
...
+func TestConnectUnixPath(t *testing.T) {
+	t.Run("Successful connection - no error", func(t *testing.T) {
+		err := testConnectUnixgram(t)
+		if err != nil {
+			t.Fatalf("Unexpected error: %v", err)
+		}
+	})
+
+	t.Run("Failed connection - End socket longer than 104 bytes", func(t *testing.T) {
+		// Retrieve HOME env variable (used by the os.UserHomeDir)
```
cfergeau, line 48: Ah right, kind of messy to do the 100 bytes test with current `main` :/ This will need to be modified once https://github.com/crc-org/vfkit/pull/195 gets in.

cfergeau (commented):
Thanks for adding this test! A few minor comments.

pkg/vf/virtionet_test.go
lstocchi: Thanks, replaced!! 👍

lstocchi:
@cfergeau updated

pkg/vf/virtionet_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+		// Assert successful execution
+		require.NoError(t, err)
+
+		// Check if path starts with the expected prefix
+		expectedPrefix := filepath.Join(userHome, "Library", "Application Support", "vfkit")
+		require.True(t, strings.HasPrefix(path, expectedPrefix), "Path doesn't start with expected prefix: %v", path)
+
+		// Check if path ends with a socket extension
+		require.Equal(t, ".sock", filepath.Ext(path), "Path doesn't end with .sock extension: %v, ext is %v", path, filepath.Ext(path))
```
anjannath, line 82: maybe we need to use `require.Equalf()/require.Truef()` variants here for the failure messages to contain the variables in the right place?

lstocchi: you're right, updated 👍

PR #201, opened 2024-10-02
contrib: update start-crc-bundle to use UEFI boot

lstocchi:
it removes all the legacy kernel related options (kernel, initrd, kernel-cmdline) and replace them with the --bootloader efi.

I found out that there is an efistore.nvram file inside the `.crc/machine.crc` path. It looks logical to use it but I'm not sure if the path can be different based on the bundle_path or anything else.

### DIFF
--- a/contrib/scripts/start-crc-bundle.sh
+++ b/contrib/scripts/start-crc-bundle.sh
@@ -4,16 +4,11 @@ set -exu
 
 YQ=${YQ:-yq}
 BUNDLE_PATH=$1
-KERNEL=$(cat ${BUNDLE_PATH}/crc-bundle-info.json | ${YQ} .nodes[0].kernel)
-INITRD=$(cat ${BUNDLE_PATH}/crc-bundle-info.json | ${YQ} .nodes[0].initramfs)
-CMDLINE=$(cat ${BUNDLE_PATH}/crc-bundle-info.json | ${YQ} .nodes[0].kernelCmdLine)
 DISKIMG=$(cat ${BUNDLE_PATH}/crc-bundle-info.json | ${YQ} .storage.diskImages[0].name)
 cp -c ${BUNDLE_PATH}/${DISKIMG} overlay.img
 
 ./out/vfkit --cpus 2 --memory 2048 \
-    --kernel "${BUNDLE_PATH}/${KERNEL}" \
-    --initrd "${BUNDLE_PATH}/${INITRD}" \
-    --kernel-cmdline "${CMDLINE}" \
+    --bootloader efi,variable-store=efistore.nvram,create \
     --device virtio-blk,path=overlay.img \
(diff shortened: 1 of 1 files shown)


lstocchi:
> Not sure about the use of `chore` in the short log, doesn't this word have a negative connotation? https://www.merriam-webster.com/dictionary/chore I think of 3. when I see it, even if I guess it's meant as 2.?

In my prev projects we used conventional commits and `chore` was related to changes that don't directly affect the application's behavior or functionality.

contrib/scripts/start-crc-bundle.sh
```diff
@@ -4,16 +4,12 @@ set -exu
...
-CMDLINE=$(cat ${BUNDLE_PATH}/crc-bundle-info.json | ${YQ} .nodes[0].kernelCmdLine)
+EFISTORE_PATH="$HOME/.crc/machines/crc/efistore.nvram"
 DISKIMG=$(cat ${BUNDLE_PATH}/crc-bundle-info.json | ${YQ} .storage.diskImages[0].name)
 cp -c ${BUNDLE_PATH}/${DISKIMG} overlay.img
 
 ./out/vfkit --cpus 2 --memory 2048 \
-    --kernel "${BUNDLE_PATH}/${KERNEL}" \
-    --initrd "${BUNDLE_PATH}/${INITRD}" \
-    --kernel-cmdline "${CMDLINE}" \
+    --bootloader efi,variable-store=${EFISTORE_PATH},create \
```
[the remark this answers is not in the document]
lstocchi: updated

lstocchi:
@cfergeau I updated the label to the commit to `contrib:` to be in line with the modifications done in the past

contrib/scripts/start-crc-bundle.sh
```diff
@@ -4,16 +4,14 @@ set -exu
...
 DISKIMG=$(cat ${BUNDLE_PATH}/crc-bundle-info.json | ${YQ} .storage.diskImages[0].name)
 cp -c ${BUNDLE_PATH}/${DISKIMG} overlay.img
+cp -c ${EFISTORE_PATH} efistore.nvram
+
 
 ./out/vfkit --cpus 2 --memory 2048 \
-    --kernel "${BUNDLE_PATH}/${KERNEL}" \
-    --initrd "${BUNDLE_PATH}/${INITRD}" \
-    --kernel-cmdline "${CMDLINE}" \
+    --bootloader efi,variable-store=efistore.nvram,create \
```
cfergeau, line 14: `create` will ensure that `efistore.nvram` is created, we don't need to reuse `"$HOME/.crc/machines/crc/efistore.nvram"` which may not exist
Using a new efi variable store will ensure we are as close as possible to a 'blank state' for our VM.

lstocchi: ah, i misunderstood your prev comment. Updated

PR #202, opened 2024-10-04
rosetta: add option to ignore rosetta failure when installation cancelled

### LINKED ISSUE #160: Add option to ignore rosetta's failure to install
When being prompted for rosetta installation, it's possible for the user to cancel the installation if they don't want rosetta. If the user does this, `vfkit` will exit with an error as rosetta was requested but it's not available.
podman would like for this behaviour to be optional, see https://github.com/containers/podman/issues/23153

lstocchi:
This adds a ignore-if-missing option that allows vfkit to not fail if the rosetta installation is cancelled by the user.
So far, if the rosetta installation was cancelled vfkit exited with an error. This behavior still exists if ignore-if-missing option is not set.
it fixes #160

### DIFF
--- a/pkg/vf/rosetta_arm64.go
+++ b/pkg/vf/rosetta_arm64.go
@@ -2,22 +2,36 @@ package vf
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/Code-Hex/vz/v3"
 	log "github.com/sirupsen/logrus"
 )
 
-func checkRosettaAvailability(install bool) error {
-	availability := vz.LinuxRosettaDirectoryShareAvailability()
+var (
+	checkRosettaDirectoryShareAvailability = vz.LinuxRosettaDirectoryShareAvailability
+	doInstallRosetta                       = vz.LinuxRosettaDirectoryShareInstallRosetta
+)
+
+func (dev *RosettaShare) checkRosettaAvailability() error {
+	availability := checkRosettaDirectoryShareAvailability()
 	switch availability {
 	case vz.LinuxRosettaAvailabilityNotSupported:
 		return fmt.Errorf("rosetta is not supported")
 	case vz.LinuxRosettaAvailabilityNotInstalled:
-		if !install {
+		if !dev.InstallRosetta {
 			return fmt.Errorf("rosetta is not installed")
 		}
 		log.Debugf("installing rosetta")
-		if err := vz.LinuxRosettaDirectoryShareInstallRosetta(); err != nil {
+		if err := doInstallRosetta(); err != nil {
+			if dev.IgnoreIfMissing {
+				log.Info("Rosetta installation failed. Continuing without Rosetta.")
+				_, err = os.Stderr.WriteString(err.Error() + "\n")
+				if err != nil {
+					log.Debugf("Failed to write error to stderr: %v", err)
+				}
+				return nil
+			}
 			return fmt.Errorf("failed to install rosetta: %w", err)
 		}
 		log.Debugf("rosetta installed")
@@ -32,7 +46,7 @@ func (dev *RosettaShare) toVz() (vz.DirectorySharingDeviceConfiguration, error) 
 	if dev.MountTag == "" {
 		return nil, fmt.Errorf("missing mandatory 'mountTage' option for rosetta share")
 	}
-	if err := checkRosettaAvailability(dev.InstallRosetta); err != nil {
+	if err := dev.checkRosettaAvailability(); err != nil {
 		return nil, err
 	}
 
--- a/pkg/vf/rosetta_arm64_test.go
+++ b/pkg/vf/rosetta_arm64_test.go
@@ -1,0 +1,110 @@
+package vf
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/Code-Hex/vz/v3"
+	"github.com/crc-org/vfkit/pkg/config"
+	"github.com/stretchr/testify/require"
+)
+
+type checkRosettaAvailabilityTest struct {
+	installRosetta                         bool
+	ignoreIfMissing                        bool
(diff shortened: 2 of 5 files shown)


pkg/vf/rosetta_arm64.go
```diff
@@ -2,22 +2,32 @@ package vf
...
 	case vz.LinuxRosettaAvailabilityNotInstalled:
-		if !install {
+		if !dev.InstallRosetta {
 			return fmt.Errorf("rosetta is not installed")
 		}
 		log.Debugf("installing rosetta")
-		if err := vz.LinuxRosettaDirectoryShareInstallRosetta(); err != nil {
+		if err := doInstallRosetta(); err != nil {
+			if dev.IgnoreIfMissing && strings.Contains(err.Error(), fmt.Sprintf("VZErrorDomain Code=%d", vz.ErrorOperationCancelled)) {
+				log.Debugf("user cancelled rosetta installation but ignoreIfMissing option is true")
```
cfergeau, line 28: Is there a way for (eg) podman to detect the rosetta installation failed? If not, we need to print something to stdout or stderr, which is not great, but will allow podman to know what happened.

pkg/vf/rosetta_arm64_test.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+	origCheckRosettaDirectoryShareAvailability := checkRosettaDirectoryShareAvailability
+	checkRosettaDirectoryShareAvailability = test.checkRosettaDirectoryShareAvailability
+	origDoInstallRosetta := doInstallRosetta
+	doInstallRosetta = test.doInstallRosetta
+	defer func() {
+		checkRosettaDirectoryShareAvailability = origCheckRosettaDirectoryShareAvailability
+		doInstallRosetta = origDoInstallRosetta
+	}()
```
cfergeau, line 101: Not sure you need to restore the initial values after running the test? I'd expect each `xxx_test.go` file to run in its own process (to be checked though ;)

cfergeau (commented):
Haven't tested this extensively, but looks good to me.
It would be nice to update the documentation as well to mention the new option, and to add a `Fixes https://github.com/crc-org/vfkit/issues/160` in the commit log so that the corresponding issue is automatically fixed.

pkg/vf/rosetta_arm64_test.go
lstocchi: So, I've been digging on it a bit and I didn't find any clear response but by looking at the code it seems that each test is run by launching a go routine https://github.com/golang/go/blob/182bdbb1e1d62fe53f8386e7b02c7738898d373b/src/testing/testing.go#L555 and from what I read all goroutines within a Go program share the same process so all tests run in the same process.
By creating a new test file I can see it to be true so I think it's better to keep the defer to avoid problems for future tests.

pkg/vf/rosetta_arm64.go
```diff
@@ -2,22 +2,32 @@ package vf
...
 		return fmt.Errorf("rosetta is not supported")
 	case vz.LinuxRosettaAvailabilityNotInstalled:
-		if !install {
+		if !dev.InstallRosetta {
 			return fmt.Errorf("rosetta is not installed")
 		}
 		log.Debugf("installing rosetta")
-		if err := vz.LinuxRosettaDirectoryShareInstallRosetta(); err != nil {
+		if err := doInstallRosetta(); err != nil {
+			if dev.IgnoreIfMissing && strings.Contains(err.Error(), fmt.Sprintf("VZErrorDomain Code=%d", vz.ErrorOperationCancelled)) {
```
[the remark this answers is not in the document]
lstocchi: I think I tried the `errors.As` without success but you made a nice observation, I agree we should ignore all errors, makes sense. Updating it

pkg/vf/rosetta_arm64.go
lstocchi: So, not sure what to do here and what's the suggested way to handle it. I just added a generic message to the stdout. Should I also print the error? I didn't print it on the stderr bc I think that if one listens to the stderr and receive something it assumes there was some kind of failure, but in this case we want to consciously skip it

lstocchi:
Forgot to update the documentation, updating it

pkg/vf/rosetta_arm64_test.go
cfergeau: `go help test` has some details on this:
> Each listed package causes the execution of a separate test binary.
so all tests in the same dir will run in parallel in go routine.
Tests in different dirs should run in separate processes.

pkg/vf/rosetta_arm64.go
```diff
@@ -2,22 +2,36 @@ package vf
...
 	case vz.LinuxRosettaAvailabilityNotInstalled:
-		if !install {
+		if !dev.InstallRosetta {
 			return fmt.Errorf("rosetta is not installed")
 		}
 		log.Debugf("installing rosetta")
-		if err := vz.LinuxRosettaDirectoryShareInstallRosetta(); err != nil {
+		if err := doInstallRosetta(); err != nil {
+			if dev.IgnoreIfMissing {
+				os.Stdout.WriteString("Rosetta installation failed. Continuing without Rosetta.")
```
cfergeau, line 28: I'd just use `log.Info("Rosetta installation failed ...")`. Logging of the error is fine to keep in debug logs (`log.Debugf`)

pkg/vf/rosetta_arm64_test.go
cfergeau: No need to make any changes though :)

pkg/vf/rosetta_arm64.go
lstocchi: Done 👍

PR #209, opened 2024-10-17
ignition: add support for ignition config file

### LINKED ISSUE #124: Add `--ignition config.ign` support
This would simplify use of generic VM images which can be configured using ignition.

The podman code doing this is https://github.com/containers/podman/blob/6487940534c1065d6c7753e3b6dfbe253666537d/pkg/machine/applehv/ignition.go

lstocchi:
it adds support to pass an ignition config file to vfkit that handles the provisioning process.

it resolves #124

### DIFF
--- a/cmd/vfkit/main.go
+++ b/cmd/vfkit/main.go
@@ -21,6 +21,9 @@ package main
 
 import (
 	"fmt"
+	"io"
+	"net"
+	"net/http"
 	"os"
 	"os/signal"
 	"runtime"
@@ -86,6 +89,10 @@ func newVMConfiguration(opts *cmdline.Options) (*config.VirtualMachine, error) {
 		return nil, err
 	}
 
+	if err := vmConfig.AddIgnitionFileFromCmdLine(opts.IgnitionPath); err != nil {
+		return nil, fmt.Errorf("failed to add ignition file: %w", err)
+	}
+
 	return vmConfig, nil
 }
 
@@ -137,6 +144,21 @@ func runVFKit(vmConfig *config.VirtualMachine, opts *cmdline.Options) error {
 }
 
 func runVirtualMachine(vmConfig *config.VirtualMachine, vm *vf.VirtualMachine) error {
+	if vm.Config().Ignition != nil {
+		go func() {
+			file, err := os.Open(vmConfig.Ignition.ConfigPath)
+			if err != nil {
+				log.Error(err)
+			}
+			defer file.Close()
+			reader := file
+			if err := startIgnitionProvisionerServer(reader, vmConfig.Ignition.SocketPath); err != nil {
+				log.Error(err)
+			}
+			log.Debug("ignition vsock server exited")
+		}()
+	}
+
 	if err := vm.Start(); err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


pkg/config/config.go
```diff
@@ -191,6 +202,39 @@ func (vm *VirtualMachine) TimeSync() *TimeSync {
...
+	if cmdlineOpts == "" {
+		return nil
+	}
+	opts := strings.Split(cmdlineOpts, ",")
+	if len(opts) != 1 {
+		return fmt.Errorf("ignition only accept one option in command line argument")
+	}
+
+	socketPath := filepath.Join(os.TempDir(), "ignition.sock")
+	dev, err := VirtioVsockNew(1024, socketPath, true)
```
baude, line 225: pop to const ?

baude: and maybe line above?

cmd/vfkit/main.go
```diff
@@ -86,6 +91,10 @@ func newVMConfiguration(opts *cmdline.Options) (*config.VirtualMachine, error) {
 		return nil, err
 	}
 
+	if err := vmConfig.AddIgnitionFileFromCmdLine(opts.IgnitionPath); err != nil {
+		return nil, err
```
baude, line 95: perhaps consider wrapping this here to make it meaningful

cmd/vfkit/main.go
```diff
@@ -198,3 +216,37 @@ func runVirtualMachine(vmConfig *config.VirtualMachine, vm *vf.VirtualMachine) e
 
 	return <-errCh
 }
+
+func startServer(ignitionFile string, ignitionSocketPath string) error {
```
baude, line 220: consider passing ignitionFile as a reader

baude:
flyby review, nice work work @lstocchi ... all my comments are largely things to think about, neat idea too.

lstocchi:
@baude thanks for the suggestions. Hopefully I got all of them. 👍 

Gonna add some tests tomorrow morning so that it can be reviewed again

baude:
@lstocchi being that ignition is on the rarer side of things, I wonder you should create `contrib/ignition` and drop something of an example in there ?  I assume it is in JSON so not something you could heavily comment on but a README.md in that dir could explain it ?  This is extra credit of course :)

baude (approved):
Just the proposal to add an example file comment ... lgtm otherwise

pkg/config/json_test.go
```diff
@@ -85,6 +85,16 @@ var jsonTests = map[string]jsonTest{
 		},
 		expectedJSON: `{"vcpus":3,"memoryBytes":4194304000,"bootloader":{"kind":"linuxBootloader","vmlinuzPath":"/vmlinuz","initrdPath":"/initrd","kernelCmdLine":"console=hvc0"},"timesync":{"vsockPort":1234}}`,
 	},
+	"TestIgnition": {
```
cfergeau, line 88: These `json_test` changes could be squashed in the previous commit.

lstocchi: updated 👍

cmd/vfkit/main.go
```diff
@@ -198,3 +220,33 @@ func runVirtualMachine(vmConfig *config.VirtualMachine, vm *vf.VirtualMachine) e
 
 	return <-errCh
 }
+
+func startIgnitionProvisionerServer(ignitionReader io.Reader, ignitionSocketPath string) error {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", func(w http.ResponseWriter, _ *http.Request) {
+		_, err := io.Copy(w, ignitionReader)
```
cfergeau, line 227: This could use https://pkg.go.dev/net/http#ServeContent (no need to make this change right, this can be a follow-up)

cfergeau (approved):
/lgtm
/approve

PR #212, opened 2024-10-30
feat: add support for NBD

### LINKED ISSUE #129: Add NBD support
This would allow for example qcow2 support through qemu-nbd https://rwmj.wordpress.com/2019/10/03/how-to-edit-a-qcow2-file-from-c/

Code-Hex/vz does not have NBD support yet, so the needed bindings would have to be written first.

See https://developer.apple.com/videos/play/wwdc2023/10007/?time=421 for a short description of NBD in virtualization framework.

lstocchi:
This patch adds support for NBD. The NetworkBlockDevice protocol allows remote block devices to be accessed over the network. By leveraging it we can connect our VM to a remote block device and use it as a local disk.

User can specify the timeout and the synchronization mode (none or full). More info at https://developer.apple.com/documentation/virtualization/vzdisksynchronizationmode?language=objc

It also listens to changes to the network block device attachment, so if there is some connectivity changes (connect, disconnect), the user is informed. To achieve it, it leverages the delegate provided by the Virtualization Framework. More info at https://developer.apple.com/documentation/virtualization/vznetworkblockdevicestoragedeviceattachmentdelegate?language=objc

This feature is only available for macOS 14.0+ and requires an update to the VZ library. So this implementation could be modified based on the work done there.

It needs https://github.com/Code-Hex/vz/pull/156 and https://github.com/Code-Hex/vz/issues/166

It resolves #129

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -178,6 +178,16 @@ func (vm *VirtualMachine) VirtioVsockDevices() []*VirtioVsock {
 	return vsockDevs
 }
 
+func (vm *VirtualMachine) NetworkBlockDevice(deviceID string) *NetworkBlockDevice {
+	for _, dev := range vm.Devices {
+		if nbdDev, isNbdDev := dev.(*NetworkBlockDevice); isNbdDev && nbdDev.DeviceIdentifier == deviceID {
+			return nbdDev
+		}
+	}
+
+	return nil
+}
+
 // AddDevice adds a dev to vm. This device can be created with one of the
 // VirtioXXXNew methods.
 func (vm *VirtualMachine) AddDevice(dev VirtioDevice) error {
--- a/pkg/config/json_test.go
+++ b/pkg/config/json_test.go
@@ -274,7 +274,7 @@ var jsonStabilityTests = map[string]jsonStabilityTest{
 			return blk
 		},
 
-		skipFields:   []string{"DevName"},
+		skipFields:   []string{"DevName", "URI"},
 		expectedJSON: `{"kind":"virtioblk","devName":"virtio-blk","imagePath":"ImagePath","readOnly":true,"deviceIdentifier":"DeviceIdentifier"}`,
 	},
 	"USBMassStorage": {
@@ -283,7 +283,7 @@ var jsonStabilityTests = map[string]jsonStabilityTest{
 			require.NoError(t, err)
 			return usb
 		},
-		skipFields:   []string{"DevName"},
+		skipFields:   []string{"DevName", "URI"},
 		expectedJSON: `{"kind":"usbmassstorage","devName":"usb-mass-storage","imagePath":"ImagePath","readOnly":true}`,
 	},
 	"NVMExpressController": {
@@ -292,7 +292,7 @@ var jsonStabilityTests = map[string]jsonStabilityTest{
 			require.NoError(t, err)
 			return nvme
 		},
-		skipFields:   []string{"DevName"},
+		skipFields:   []string{"DevName", "URI"},
 		expectedJSON: `{"kind":"nvme","devName":"nvme","imagePath":"ImagePath","readOnly":true}`,
 	},
 	"LinuxBootloader": {
--- a/pkg/config/json_test.go
+++ b/pkg/config/json_test.go
@@ -170,12 +170,15 @@ var jsonTests = map[string]jsonTest{
 			// rosetta
 			rosetta, err := RosettaShareNew("vz-rosetta")
 			require.NoError(t, err)
-			err = vm.AddDevices(fs, usb, rosetta)
+			// NBD
+			nbd, err := NetworkBlockDeviceNew("uri", 1, SynchronizationFullMode)
+			require.NoError(t, err)
+			err = vm.AddDevices(fs, usb, rosetta, nbd)
 			require.NoError(t, err)
 
 			return vm
 		},
-		expectedJSON: `{"vcpus":3,"memoryBytes":4194304000,"bootloader":{"kind":"linuxBootloader","vmlinuzPath":"/vmlinuz","initrdPath":"/initrd","kernelCmdLine":"console=hvc0"},"devices":[{"kind":"virtioserial","logFile":"/virtioserial"},{"kind":"virtioinput","inputType":"keyboard"},{"kind":"virtiogpu","usesGUI":false,"width":800,"height":600},{"kind":"virtionet","nat":true,"macAddress":"00:11:22:33:44:55"},{"kind":"virtiorng"},{"kind":"virtioblk","devName":"virtio-blk","imagePath":"/virtioblk"},{"kind":"virtiosock","port":1234,"socketURL":"/virtiovsock"},{"kind":"virtiofs","mountTag":"tag","sharedDir":"/virtiofs"},{"kind":"usbmassstorage","devName":"usb-mass-storage","imagePath":"/usbmassstorage","readOnly":true},{"kind":"rosetta","mountTag":"vz-rosetta","installRosetta":false,"ignoreIfMissing":false}]}`,
+		expectedJSON: `{"vcpus":3,"memoryBytes":4194304000,"bootloader":{"kind":"linuxBootloader","vmlinuzPath":"/vmlinuz","initrdPath":"/initrd","kernelCmdLine":"console=hvc0"},"devices":[{"kind":"virtioserial","logFile":"/virtioserial"},{"kind":"virtioinput","inputType":"keyboard"},{"kind":"virtiogpu","usesGUI":false,"width":800,"height":600},{"kind":"virtionet","nat":true,"macAddress":"00:11:22:33:44:55"},{"kind":"virtiorng"},{"kind":"virtioblk","devName":"virtio-blk","imagePath":"/virtioblk"},{"kind":"virtiosock","port":1234,"socketURL":"/virtiovsock"},{"kind":"virtiofs","mountTag":"tag","sharedDir":"/virtiofs"},{"kind":"usbmassstorage","devName":"usb-mass-storage","imagePath":"/usbmassstorage","readOnly":true},{"kind":"rosetta","mountTag":"vz-rosetta","installRosetta":false,"ignoreIfMissing":false},{"kind":"nbd", "devName":"nbd", "uri":"uri", "SynchronizationMode":"full","Timeout":1000000}]}`,
 	},
 }
 
@@ -306,6 +309,15 @@ var jsonStabilityTests = map[string]jsonStabilityTest{
 	"TimeSync": {
 		obj:          &TimeSync{},
 		expectedJSON: `{"vsockPort":3}`,
+	},
+	"NetworkBlockDevice": {
+		newObjectFunc: func(t *testing.T) any {
+			nbd, err := NetworkBlockDeviceNew("uri", 1000, SynchronizationFullMode)
+			require.NoError(t, err)
+			return nbd
+		},
+		skipFields:   []string{"DevName", "ImagePath"},
+		expectedJSON: `{"kind":"nbd","deviceIdentifier":"DeviceIdentifier","devName":"nbd","uri":"URI","readOnly":true,"SynchronizationMode":"SynchronizationMode","Timeout":2}`,
 	},
 }
 
--- a/pkg/config/virtio.go
+++ b/pkg/config/virtio.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 )
 
 // The VirtioDevice interface is an interface which is implemented by all virtio devices.
@@ -108,6 +109,19 @@ type VirtioSerial struct {
 	UsesStdio bool   `json:"usesStdio,omitempty"`
 }
 
+type NBDSynchronizationMode string
+
+const (
+	SynchronizationFullMode NBDSynchronizationMode = "full"
+	SynchronizationNoneMode NBDSynchronizationMode = "none"
+)
+
+type NetworkBlockDevice struct {
+	VirtioBlk
+	Timeout             time.Duration
+	SynchronizationMode NBDSynchronizationMode
+}
+
 // TODO: Add VirtioBalloon
 // https://github.com/Code-Hex/vz/blob/master/memory_balloon.go
 
@@ -170,6 +184,8 @@ func deviceFromCmdLine(deviceOpts string) (VirtioDevice, error) {
 		dev = &VirtioInput{}
 	case "virtio-gpu":
 		dev = &VirtioGPU{}
+	case "nbd":
+		dev = networkBlockDeviceNewEmpty()
 	default:
 		return nil, fmt.Errorf("unknown device type: %s", opts[0])
 	}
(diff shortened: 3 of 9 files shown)


lstocchi:
Added initial tests and documentation. I plan to also add an e2e test.
I'll wait the vz patch to be merged before refining everything but ready for an early review if you want

lstocchi:
> Added initial tests and documentation. I plan to also add an e2e test. I'll wait the vz patch to be merged before refining everything but ready for an early review if you want

After looking again, I'll skip the e2e test bc I need Mac 14 to work and the current workflow skip it, so maybe we can work on it in a separate PR if we want to do it.

For the rest it should be ready for a review. We just need to merge https://github.com/crc-org/vfkit/pull/214 and rebase it.
@cfergeau

pkg/vf/virtio.go
```diff
@@ -289,6 +290,73 @@ func (dev *VirtioVsock) AddToVirtualMachineConfig(vmConfig *VirtualMachineConfig
 	return nil
 }
 
+func (dev *NetworkBlockDevice) toVz() (vz.StorageDeviceConfiguration, error) {
+	if dev.Uri == "" || dev.DeviceIdentifier == "" {
+		return nil, fmt.Errorf("missing required options for NBD device: 'uri' and 'deviceId' must be specified")
```
baude, line 295: here you check for empty options, but do you ever check to see if they are valid?  If you do and you confident in that, maybe move this to that section?

pkg/vf/virtio.go
```diff
@@ -289,6 +290,73 @@ func (dev *VirtioVsock) AddToVirtualMachineConfig(vmConfig *VirtualMachineConfig
...
+	if err != nil {
+		return nil, err
+	}
+	vzdev.SetBlockDeviceIdentifier(dev.DeviceIdentifier)
+
+	return vzdev, nil
+}
+
+func (dev *NetworkBlockDevice) SyncronizationModeVZ() vz.DiskSynchronizationMode {
+	switch dev.SynchronizationMode {
```
baude, line 313: if with an early return and then return the default ?

lstocchi:
@baude updated 👍

lstocchi:
Forgot to mark it ready 😄

pkg/vf/virtio.go
```diff
@@ -289,6 +292,112 @@ func (dev *VirtioVsock) AddToVirtualMachineConfig(vmConfig *VirtualMachineConfig
...
+		return fmt.Errorf("'uri' must be specified")
+	}
+
+	parsed, err := url.Parse(uri)
+	if err != nil {
+		return fmt.Errorf("error: %w", err)
+	}
+
+	// The format specified by https://github.com/NetworkBlockDevice/nbd/blob/master/doc/uri.md
+	if parsed.Scheme != "nbd" && parsed.Scheme != "nbds" && parsed.Scheme != "nbd+unix" && parsed.Scheme != "nbds+unix" {
```
cfergeau, line 332: `NetworkBlockDeviceNew` in `pkg/config/virtio.go` says:
```
// The provided uri must be in the format nbd://<address>/<export-name>.
```
but we parse more than just `nbd://`. The https://github.com/NetworkBlockDevice/nbd/blob/master/doc/uri.md link would be useful to have in the doc for `NetworkBlockDeviceNew`

pkg/vf/virtio.go
```diff
@@ -289,6 +292,112 @@ func (dev *VirtioVsock) AddToVirtualMachineConfig(vmConfig *VirtualMachineConfig
...
+	return nil
+}
+
+func (dev *NetworkBlockDevice) validateNbdDeviceIdentifier(deviceID string) error {
+	if deviceID == "" {
+		return fmt.Errorf("'deviceId' must be specified")
+	}
+
+	if strings.Contains(deviceID, "/") {
+		return fmt.Errorf("invalid 'deviceId': it cannot container any forward slash")
```
cfergeau, line 345: 'contain'

pkg/config/json_test.go
```diff
@@ -274,7 +274,7 @@ var jsonStabilityTests = map[string]jsonStabilityTest{
 			return blk
 		},
 
-		skipFields:   []string{"DevName"},
+		skipFields:   []string{"DevName", "Uri"},
```
cfergeau: I'm surprised by the `Uri` capitalization, it is `URI` in the struct, do we need to skip this `Uri` field?

pkg/config/config.go
```diff
@@ -178,6 +178,16 @@ func (vm *VirtualMachine) VirtioVsockDevices() []*VirtioVsock {
 	return vsockDevs
 }
 
+func (vm *VirtualMachine) NetworkBlockDevice(deviceID string) *NetworkBlockDevice {
```
cfergeau, line 181: Is there a limit of 1 network block device per-VM?

lstocchi: No you can have more. This is used when we want to retrieve info about one specific nbd from https://github.com/crc-org/vfkit/pull/212/files#diff-04ef060002636f3776175738389d7448ca395b65d0db23da5137ade7c279a418R382

pkg/config/json_test.go
lstocchi: Maybe you were looking at an old version? I can see it is actually URI in the code.

pkg/vf/virtio.go
lstocchi: Right. Updating 👍

lstocchi: Updated the comment and the doc 👍

pkg/config/json_test.go
cfergeau: Confused me when looking again, but the first commit adds `Uri`, the second commit changes this to `URI`

pkg/config/config.go
cfergeau: Ah oops, did not pay attention to the `deviceId` parameter

pkg/config/json_test.go
lstocchi: Ah i see. At first it was Uri, then lint complained but it seems I didn't fix all commits 👍

lstocchi: Updated, sorry I didn't see it before. Thanks!!

pkg/config/virtio.go
```diff
@@ -663,6 +679,76 @@ func (dev *RosettaShare) FromOptions(options []option) error {
...
 
+func networkBlockDeviceNewEmpty() *NetworkBlockDevice {
+	return &NetworkBlockDevice{
+		VirtioBlk: VirtioBlk{
+			StorageConfig: StorageConfig{
+				DevName: "nbd",
+			},
+			DeviceIdentifier: "",
+		},
+		Timeout:             time.Duration(15000 * time.Millisecond), // set a default timeout to 15s
```
cfergeau, line 690: Does this have to be set to a default value? Or can we use the virtualization framework's default if the user did not explicitly set this timeout?

pkg/config/virtio.go
```diff
@@ -663,6 +679,76 @@ func (dev *RosettaShare) FromOptions(options []option) error {
...
+
+	return nbd, nil
+}
+
+func (nbd *NetworkBlockDevice) ToCmdLine() ([]string, error) {
+	cmdLine, err := nbd.VirtioBlk.ToCmdLine()
+	if err != nil {
+		return []string{}, err
+	}
+
```
cfergeau, line 714: I'd add the same check as in `VirtioBlk.ToCmdLine()`:
```
	if len(cmdLine) != 2 {
		return []string{}, fmt.Errorf("unexpected storage config commandline")
	}
```

pkg/config/virtio.go
```diff
@@ -691,15 +777,26 @@ func (dev *USBMassStorage) SetReadOnly(readOnly bool) {
...
 
 func (config *StorageConfig) ToCmdLine() ([]string, error) {
-	if config.ImagePath == "" {
-		return nil, fmt.Errorf("%s devices need the path to a disk image", config.DevName)
+	if config.ImagePath != "" && config.URI != "" {
+		return nil, fmt.Errorf("%s devices cannot have both path to a disk image and a uri to a remote block device", config.DevName)
+	}
+	if config.ImagePath == "" && config.URI == "" {
+		return nil, fmt.Errorf("%s devices need a path to a disk image or a uri to a remote block device", config.DevName)
+	}
```
cfergeau, line 791: Only a comment (or just me thinking out loud), I'm not asking for changes in response to this comment.
We could having both `ImagePath` and `URI` in `StorageConfig`, and avoid these checks with something like:

```
type StorageConfig struct {
	DevName   string `json:"devName"`
	ReadOnly  bool   `json:"readOnly,omitempty"`
}

type DiskStorageConfig {
    StorageConfig
    ImagePath string `json:"imagePath,omitempty"`
}

type NetworkBlockDevice {
    StorageConfig
    URI       string `json:"uri,omitempty"`
}
```

If it works, this makes sure we'll never get a NetworkBlockDevice with ImagePath set as it does not have this member at all.
This adds one more almost empty type though, and may be breaking unserializing json on upgrades (?).

pkg/vf/virtio.go
```diff
@@ -289,6 +292,112 @@ func (dev *VirtioVsock) AddToVirtualMachineConfig(vmConfig *VirtualMachineConfig
...
+}
+
+func ListenNetworkBlockDevices(vm *VirtualMachine) error {
+	for _, dev := range vm.vfConfig.storageDevicesConfiguration {
+		attachment := dev.Attachment()
+		if nbdAttachment, isNbdAttachment := attachment.(*vz.NetworkBlockDeviceStorageDeviceAttachment); isNbdAttachment {
+			nbdID, err := dev.(*vz.VirtioBlockDeviceConfiguration).BlockDeviceIdentifier()
+			if err != nil {
+				log.Info("Found NBD device with no identifier. Please set a deviceId to each NBD device.")
+				return fmt.Errorf("NetworkBlockDevice must have deviceId. Error %v", err)
```
cfergeau, line 380: you could use `%w` instead of `%v` so that the error is wrapped in the returned error object instead of only `err.String()` being added to the error string.

pkg/vf/virtio.go
```diff
@@ -289,6 +292,112 @@ func (dev *VirtioVsock) AddToVirtualMachineConfig(vmConfig *VirtualMachineConfig
...
+			}
+			nbdConfig := vm.vfConfig.config.NetworkBlockDevice(nbdID)
+			if nbdConfig == nil {
```
cfergeau, line 383: We go through quite a few steps (several interface -> concrete type conversions, one iteration over the VM devices to find the right ones, ...) because we want to print `nbdConfig.URI` in our log messages.

This `URI` is available directly from the `nbdAttachment` we get early, but the API for that is not implemented in Code-Hex/vz: https://developer.apple.com/documentation/virtualization/vznetworkblockdevicestoragedeviceattachment/4168501-url?language=objc
Maybe we could only print the nbd device ID in the logs for now, and file a bug or add the API to Code-Hex/vz, and then update this code? This would result in simpler code.

cfergeau: An alternative with slightly simpler code is this:
```
diff --git a/pkg/vf/virtio.go b/pkg/vf/virtio.go
index 39af346..c2e78ef 100644
--- a/pkg/vf/virtio.go
+++ b/pkg/vf/virtio.go
@@ -29,6 +29,11 @@ type VirtioInput config.VirtioInput
 type VirtioGPU config.VirtioGPU
 type NetworkBlockDevice config.NetworkBlockDevice
 
+type vzNetworkBlockDevice struct {
+       *vz.VirtioBlockDeviceConfiguration
+       config *NetworkBlockDevice
+}
+
 func (dev *NVMExpressController) toVz() (vz.StorageDeviceConfiguration, error) {
        var storageConfig StorageConfig = StorageConfig(dev.StorageConfig)
        attachment, err := storageConfig.toVz()
@@ -315,7 +320,7 @@ func (dev *NetworkBlockDevice) toVz() (vz.StorageDeviceConfiguration, error) {
                return nil, err
        }
 
-       return vzdev, nil
+       return vzNetworkBlockDevice{VirtioBlockDeviceConfiguration: vzdev, config: dev}, nil
 }
 
 func (dev *NetworkBlockDevice) validateNbdURI(uri string) error {
@@ -372,17 +377,14 @@ func (dev *NetworkBlockDevice) AddToVirtualMachineConfig(vmConfig *VirtualMachin
 
 func ListenNetworkBlockDevices(vm *VirtualMachine) error {
        for _, dev := range vm.vfConfig.storageDevicesConfiguration {
-               attachment := dev.Attachment()
-               if nbdAttachment, isNbdAttachment := attachment.(*vz.NetworkBlockDeviceStorageDeviceAttachment); isNbdAttachment {
-                       nbdID, err := dev.(*vz.VirtioBlockDeviceConfiguration).BlockDeviceIdentifier()
-                       if err != nil {
-                               log.Info("Found NBD device with no identifier. Please set a deviceId to each NBD device.")
-                               return fmt.Errorf("NetworkBlockDevice must have deviceId. Error %v", err)
-                       }
-                       nbdConfig := vm.vfConfig.config.NetworkBlockDevice(nbdID)
-                       if nbdConfig == nil {
-                               return fmt.Errorf("no NBD device found with deviceId %s", nbdID)
+               if nbdDev, isNbdDev := dev.(vzNetworkBlockDevice); isNbdDev {
+                       nbdAttachment, isNbdAttachment := dev.Attachment().(*vz.NetworkBlockDeviceStorageDeviceAttachment)
+                       if !isNbdAttachment {
+                               log.Info("Found NBD device with no NBD attachment. Please file a vfkit bug.")
+                               return fmt.Errorf("NetworkBlockDevice must use a NBD attachment")
                        }
+
+                       nbdConfig := nbdDev.config
                        go func() {
                                for {
                                        select {
```

I've only compile-tested it, so I'm not 100% sure it works.

cfergeau: Overall, your call, I'm fine with your initial code if you prefer to keep it this way.

pkg/config/virtio.go
lstocchi: I think I didn't add it bc in the row above it calls `cmdLine, err := nbd.VirtioBlk.ToCmdLine()` so the check is already performed in the `VirtioBlk.ToCmdLine()`. Do you prefer to have it duplicated?

cfergeau: In my opinion it is better to have it here as well. If one day we refactor this code and it no longer calls `VirtioBlk.ToCmdLine()` we will lose that check without noticing.

pkg/config/virtio.go
lstocchi: I like this idea but I propose to work on it in a follow-up PR. I see that StorageConfig is linked to many other things like NVMExpressController, VirtioBlk... I think it's better to have this refactoring separated for an easier review/testing

lstocchi: I'll open an issue if you agree

cfergeau: Yes, a follow-up issue is good, as I said:
> Only a comment (or just me thinking out loud)

pkg/vf/virtio.go
```diff
@@ -289,6 +292,112 @@ func (dev *VirtioVsock) AddToVirtualMachineConfig(vmConfig *VirtualMachineConfig
...
+	}
+
+	if len(deviceID) > 255 {
+		return fmt.Errorf("invalid 'deviceId': exceeds maximum length")
+	}
+
+	return nil
+}
+
+func (dev *NetworkBlockDevice) SyncronizationModeVZ() vz.DiskSynchronizationMode {
```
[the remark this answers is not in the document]
lstocchi: done 👍

pkg/vf/virtio.go
lstocchi: Applied your suggestion 👍

cfergeau (approved):

Issue #174, opened 2024-08-12
Surprising values returned for the restful "state"


afbjorklund:
https://github.com/crc-org/vfkit/blob/main/doc/usage.md#restful-service

> Get VM state
> Used to obtain the state of the virtual machine that is being run by vfkit.
>
> `GET /vm/state` Response: { "state": string, "canStart": bool, "canPause": bool, "canResume": bool, "canStop": bool, "canHardStop": bool 

It doesn't show the possible values of "string", but it doesn't match the POST:

```json
{
  "canHardStop": true,
  "canPause": true,
  "canResume": false,
  "canStart": false,
  "canStop": true,
  "state": "VirtualMachineStateRunning"
}
```

It should probably be documented? I think the values come straight from vz.

cfergeau:
They are indeed coming from https://github.com/Code-Hex/vz/blob/c3198942b1b797ac424b673e06769f45bef0d275/virtualmachinestate_string.go#L22C1-L23C93

I agree it would make more sense to strip the `VirtualMachineState`, I don't think there are many users of this API right now (I'll have to check) so we can probably just do it.
And yes, documenting the possible values would be useful, a list is available there: https://github.com/Code-Hex/vz/blob/c3198942b1b797ac424b673e06769f45bef0d275/virtualization.go#L25-L67

afbjorklund:
I used both types of values
(both "enum" and string), but it did raise some eyebrows during code review...

afbjorklund:
Documenting them might be easiest

cfergeau:
> Documenting them might be easiest

Yes, it seems it is too late to be changing them, podman is parsing the `VirtualMachineState*` values, `krunkit` exposes a REST API with these values https://github.com/containers/krunkit/blob/main/docs/usage.md#restful-service , ...

Better to add documentation as you suggested.

afbjorklund:
https://en.wikipedia.org/wiki/Hyrum%27s_Law

cfergeau:
If/when we make a breaking `v2` API, we can fix these suboptimal names.

PR #224, opened 2024-11-25
doc: Document the 'state' value in `GET /vm/status`

cfergeau:
This fixes https://github.com/crc-org/vfkit/issues/174


lstocchi (approved):
LGTM

PR #259, opened 2025-02-14
serial: Add --device virtio-serial,pty support, with ConsoleDeviceConfiguration

evidolob:
This PR based on https://github.com/crc-org/vfkit/pull/113
Fixes https://github.com/crc-org/vfkit/issues/48

As @cfergeau suggested PR set 'isConsole' for [VZVirtioConsolePortConfiguration](https://developer.apple.com/documentation/virtualization/vzvirtioconsoleportconfiguration?language=objc)

Also I found that not deleting `slave` file descriptor is increase stability of tty interface, and making connection to tty possible at any time.

And for, some, reason I found that using `minicom` instead of `screen` is more reliable on tty interaction.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -300,11 +300,12 @@ This is useful in combination with usermode networking stacks such as [gvisor-ta
 #### Description
 
 The `--device virtio-serial` option adds a serial device to the virtual machine. This is useful to redirect text output from the virtual machine to a log file.
-The `logFilePath` and `stdio` arguments are mutually exclusive.
+The `logFilePath`, `stdio`, `pty` arguments are mutually exclusive.
 
 #### Arguments
 - `logFilePath`: path where the serial port output should be written.
 - `stdio`: uses stdin/stdout for the serial console input/output.
+- `pty`: allocates a pseudo-terminal for the serial console input/output.
 
 #### Example
 
@@ -319,6 +320,18 @@ launched from will be used as an interactive serial console for that device:
 --device virtio-serial,stdio
 ```
 
+This adds a virtio-serial device to the VM, and creates a pseudo-terminal for
+the console for that device:
+```
+--device virtio-serial,pty
+```
+Once the VM is running, you can connect to its console with:
+```
+screen /dev/ttys002
+```
+`/dev/ttys002` will vary between `vfkit` runs.
+The `/dev/ttys???` path to the pty is printed during vfkit startup.
+It's also available through the `/vm/inspect` endpoint of [REST API](#restful-service) in the `ptyName` field of the `virtio-serial` device.
 
 ### Random Number Generator
 
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -300,7 +300,7 @@ This is useful in combination with usermode networking stacks such as [gvisor-ta
 #### Description
 
 The `--device virtio-serial` option adds a serial device to the virtual machine. This is useful to redirect text output from the virtual machine to a log file.
-The `logFilePath`, `stdio`, `pty` arguments are mutually exclusive.
+The `logFilePath`, `stdio` and `pty` arguments are mutually exclusive.
 
 #### Arguments
 - `logFilePath`: path where the serial port output should be written.
--- a/pkg/config/virtio.go
+++ b/pkg/config/virtio.go
@@ -107,6 +107,10 @@ type VirtioNet struct {
 type VirtioSerial struct {
 	LogFile   string `json:"logFile,omitempty"`
 	UsesStdio bool   `json:"usesStdio,omitempty"`
+	UsesPty   bool   `json:"usesPty,omitempty"`
+	// PtyName must not be set when creating the VM, from a user perspective, it's read-only,
+	// vfkit will set it during VM startup.
+	PtyName string `json:"ptyName,omitempty"`
 }
 
 type NBDSynchronizationMode string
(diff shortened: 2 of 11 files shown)


evidolob:
@praveenkumar @anjannath It would be grate if you test this PR on your workflow

evidolob:
@cfergeau Tests are failing because of changing vm behaviour when I add console device. On test we launch vm with:
```bash
./out/vfkit \
    --cpus 2 --memory 2048 \
    --kernel /var/folders/p3/pnj31p35351bnyj6vj8zhq1w0000gn/T/TestVirtioSerialPTY393398295/001/bzImage \
--initrd /var/folders/p3/pnj31p35351bnyj6vj8zhq1w0000gn/T/TestVirtioSerialPTY393398295/001/initramfs.cpio.gz \
--kernel-cmdline console=hvc0 \
--device virtio-net,nat,mac=56:46:4b:49:54:01 \
--device virtio-serial,pty
```
But with this PR changes linux not finished networking setup until someone not connect to tty interface, so test which wait for ssh connection is exiting with timeout.

I found that removing `console=hvc0` from `--kernel-cmdline` is fixing that behaviour, but I'm not sure that it is good to commit that change.
WDYT?

pkg/vf/virtio.go
```diff
@@ -249,17 +251,38 @@ func setRawMode(f *os.File) error {
...
+		if err != nil {
+			return nil, err
+		}
+		// as far as I can tell, we have no use for the slave fd in the
+		// vfkit process, the user will open minicom/screen/... /dev/ttys00?
+		// when needed
+		defer slave.Close()
+
+		// the master fd must stay open for vfkit's lifetime
+		gocleanup.Register(func() { _ = master.Close() })
```
cfergeau: I assume this can use `util.RegisterExitHandler`?

cfergeau:
> I found that removing `console=hvc0` from `--kernel-cmdline` is fixing that behaviour, but I'm not sure that it is good to commit that change.
> WDYT?

In my opinion this is problematic, the expected behaviour is that when you use the --pty option, the VM runs in the background regardless of its configuration, and you can connect to it using minicom/screen when you need to. I don't remember seeing this behaviour, does it appear with your changes?

Is there anything shown on the GUI window when using --gui? or maybe this prevents this "freeze" behaviour from occurring?

I found https://github.com/utmapp/UTM/issues/5143 which is very similar, but no solution there :(

baude:
when we set the console in grub, do we get early boot messages now or not until we get to later in init ?

doc/usage.md
```diff
@@ -279,11 +279,12 @@ This is useful in combination with usermode networking stacks such as [gvisor-ta
 #### Description
 
 The `--device virtio-serial` option adds a serial device to the virtual machine. This is useful to redirect text output from the virtual machine to a log file.
-The `logFilePath` and `stdio` arguments are mutually exclusive.
+The `logFilePath`, `stdio`, `pty` arguments are mutually exclusive.
```
baude, line 282:
```suggestion
The `logFilePath`, `stdio`, and `pty` arguments are mutually exclusive.
```

pkg/vf/virtio.go
```diff
@@ -230,39 +231,64 @@ func unixFd(fd uintptr) int {
 // https://developer.apple.com/documentation/virtualization/running_linux_in_a_virtual_machine#3880009
 func setRawMode(f *os.File) error {
 	// Get settings for terminal
-	attr, _ := unix.IoctlGetTermios(unixFd(f.Fd()), unix.TIOCGETA)
+	var attr unix.Termios
+	err := termios.Tcgetattr(f.Fd(), &attr)
```
baude, line 235: if and only if you repush, because attr is being used as a pointer, this can be the `if err :=` syntax (i believe)

evidolob:
@cfergeau I have made a few test around this PR and launching VM with `efi` bootloader, everything works(networking) fine, even without tty connection.

It seems that when we start VM with
```bash
./out/vfkit \
    --cpus 2 --memory 2048 \
    --kernel /var/folders/p3/pnj31p35351bnyj6vj8zhq1w0000gn/T/TestVirtioSerialPTY393398295/001/bzImage \
--initrd /var/folders/p3/pnj31p35351bnyj6vj8zhq1w0000gn/T/TestVirtioSerialPTY393398295/001/initramfs.cpio.gz \
--kernel-cmdline console=hvc0 \
--device virtio-net,nat,mac=56:46:4b:49:54:01 \
--device virtio-serial,pty
```

we pass `kernel-cmdline console=hvc0`, VZ framework(or linux kernel) expects that 'somebody' connects to tty, end pauses VM until it happens.

pkg/config/virtio.go
```diff
@@ -245,11 +261,16 @@ func (dev *VirtioSerial) ToCmdLine() ([]string, error) {
 	if err := dev.validate(); err != nil {
 		return nil, err
 	}
-	if dev.UsesStdio {
+	switch {
+	case dev.UsesStdio:
 		return []string{"--device", "virtio-serial,stdio"}, nil
+	case dev.UsesPty:
+		return []string{"--device", "virtio-serial,pty"}, nil
+	case dev.LogFile != "":
```
lstocchi, line 269: is this case needed? Fallthrough makes the default case to be executed so can this be removed?

lstocchi (commented):
I've been trying with fedora coreos and rhel. The pty works with both without any specific settings from my end.

The problem that I saw is that it works fine and it's quick at replying the first time. If I turns off the terminal with the screen process and I start a new one to simulate a logoff/login it is terribly slow. Not sure why. We can also merge as it is imo, and try to enhance it further later

evidolob:
Before merge, we need to decide what we need to do with tests, as with this PR they are failing due to:
> But with this PR changes linux not finished networking setup until someone not connect to tty interface, so test which wait for ssh connection is exiting with timeout.

test/osprovider.go
```diff
@@ -193,7 +193,7 @@ func (puipui *PuiPuiProvider) Fetch(destDir string) error {
 		return err
 	}
 	log.Infof("puipui initramfs: %s", puipui.initramfs)
-	puipui.kernelArgs = "console=hvc0"
+	puipui.kernelArgs = "console=quiet"
```
cfergeau, line 196: Not sure if `console=quiet` can have adverse side-effects as `console=` is parsed by the kernel, and `quiet` is most likely an invalid name for a console. I used `quiet` (without `console=`) in my suggestion as this is a valid kernel option, and vfkit’s code did not like to have an empty kernel commandline. I’d guess `dummy` would also work.
I’d prefer we don’t use `console=quiet` as this could be confusing (unless this is a valid `console` option).
This could go in a separate commit before this one so that the commit log can have an explanation *why* we make that change.

evidolob: Yes, sorry, it was my mistake, fixed.

cfergeau (approved):

PR #263, opened 2025-02-24
allow to pass user/meta data files for cloud-init config

lstocchi:
this patch allows vfkit to accept raw cloud-init config files to customize the VM.
It uses those files to generate an ISO image and inject it using virtio-blk device (https://github.com/crc-org/vfkit/commit/0ae085d1f82069108d39f8603435a626cbd46e51)

The supported and expected files are user-data and meta-data.
As user-data can be written in different formats (https://cloudinit.readthedocs.io/en/latest/explanation/format.html)
vfkit relies on the file names to detect them.
If the user do not pass user-data or meta-data, an empty file with that name will be added into the ISO image.

Vfkit deletes the cloud-init ISO file when it exits

### DIFF
--- a/cmd/vfkit/main.go
+++ b/cmd/vfkit/main.go
@@ -20,12 +20,14 @@ limitations under the License.
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"net"
 	"net/http"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"runtime"
 	"syscall"
 	"time"
@@ -36,6 +38,7 @@ import (
 	"github.com/crc-org/vfkit/pkg/rest"
 	restvf "github.com/crc-org/vfkit/pkg/rest/vf"
 	"github.com/crc-org/vfkit/pkg/vf"
+	"github.com/kdomanski/iso9660"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/crc-org/vfkit/pkg/util"
@@ -85,6 +88,16 @@ func newVMConfiguration(opts *cmdline.Options) (*config.VirtualMachine, error) {
 
 	if err := vmConfig.AddTimeSyncFromCmdLine(opts.TimeSync); err != nil {
 		return nil, err
+	}
+
+	cloudInitISO, err := generateCloudInitImage(opts.CloudInitFiles.GetSlice())
+	if err != nil {
+		return nil, err
+	}
+
+	// if it generated a valid cloudinit config ISO file we add it to the devices
+	if cloudInitISO != "" {
+		opts.Devices = append(opts.Devices, fmt.Sprintf("virtio-blk,path=%s", cloudInitISO))
 	}
 
 	if err := vmConfig.AddDevicesFromCmdLine(opts.Devices); err != nil {
@@ -263,4 +276,93 @@ func startIgnitionProvisionerServer(ignitionReader io.Reader, ignitionSocketPath
 
 	log.Debugf("ignition socket: %s", ignitionSocketPath)
 	return srv.Serve(listener)
+}
+
+// it generates a cloud init image by taking the files passed by the user
+// as cloud-init expects files with a specific name (e.g user-data, meta-data) we check the filenames to retrieve the correct info
+// if some file is not passed by the user, an empty file will be copied to the cloud-init ISO to
+// guarantee it to work (user-data and meta-data files are mandatory and both must exists, even if they are empty)
+// if both files are missing it returns an error
+func generateCloudInitImage(files []string) (string, error) {
+	if len(files) == 0 {
+		return "", nil
+	}
+
+	configFiles := map[string]io.Reader{
+		"user-data": nil,
+		"meta-data": nil,
+	}
+
+	hasConfigFile := false
+	for _, path := range files {
+		if path == "" {
+			continue
+		}
+		file, err := os.Open(path)
+		if err != nil {
+			return "", err
+		}
+		defer file.Close()
+
+		filename := filepath.Base(path)
+		if _, ok := configFiles[filename]; ok {
+			hasConfigFile = true
+			configFiles[filename] = file
+		}
+	}
+
+	if !hasConfigFile {
+		return "", fmt.Errorf("cloud-init needs user-data and meta-data files to work")
+	}
+
+	return createCloudInitISO(configFiles)
+}
+
+// It generates a temp ISO file containing the files passed by the user
+// It also register an exit handler to delete the file when vfkit exits
+func createCloudInitISO(files map[string]io.Reader) (string, error) {
+	writer, err := iso9660.NewWriter()
+	if err != nil {
+		return "", fmt.Errorf("failed to create writer: %w", err)
+	}
+
+	defer func() {
+		if err := writer.Cleanup(); err != nil {
+			log.Error(err)
+		}
+	}()
+
+	for name, reader := range files {
+		// if reader is nil, we set it to an empty file
+		if reader == nil {
+			reader = bytes.NewReader([]byte{})
+		}
+		err = writer.AddFile(reader, name)
+		if err != nil {
+			return "", fmt.Errorf("failed to add %s file: %w", name, err)
+		}
+	}
+
+	isoFile, err := os.CreateTemp("", "vfkit-cloudinit-")
+	if err != nil {
+		return "", fmt.Errorf("unable to create temporary cloud-init ISO file: %w", err)
+	}
+
+	defer func() {
+		if err := isoFile.Close(); err != nil {
+			log.Error(fmt.Errorf("failed to close cloud-init ISO file: %w", err))
+		}
+	}()
+
+	// register handler to remove isoFile when exiting
+	util.RegisterExitHandler(func() {
+		os.Remove(isoFile.Name())
+	})
+
+	err = writer.WriteTo(isoFile, "cidata")
+	if err != nil {
+		return "", fmt.Errorf("failed to write cloud-init ISO image: %w", err)
+	}
+
+	return isoFile.Name(), nil
 }
--- a/cmd/vfkit/main_test.go
+++ b/cmd/vfkit/main_test.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"os"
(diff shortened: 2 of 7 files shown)


cmd/vfkit/main.go
```diff
@@ -264,3 +278,86 @@ func startIgnitionProvisionerServer(ignitionReader io.Reader, ignitionSocketPath
...
+			return "", err
+		}
+		defer file.Close()
+
+		data, err := io.ReadAll(file)
+		if err != nil {
+			return "", err
+		}
+
+		// based on the file name we save its content on the corresponding var
```
baude, line 303: is it possible to  use marshalling to determine the type?

cmd/vfkit/main.go
```diff
@@ -264,3 +278,86 @@ func startIgnitionProvisionerServer(ignitionReader io.Reader, ignitionSocketPath
...
+	if err != nil {
+		return "", errors.Wrap(err, fmt.Sprintf("failed to add file: %s", err))
+	}
+
+	cacheDir, err := os.UserCacheDir()
+	if err != nil {
+		return "", errors.Wrap(err, "unable to fetch user cache dir")
+	}
+
+	isoFilePath := filepath.Join(cacheDir, "vfkit-cloudinit.iso")
```
baude, line 341: is this file deleted after vfkit exits?  If not, I wonder if that filename should be somehow different per run? I dont know the origin of cachedir, so if the runs are namespaced, then you are probably cool.  But I was thinking of a scneario where two vms are run, both with cloud-init, would they overwrite eachother?

baude:
really cool idea ... i like where you are taking this.  couple of questions from me bc I had to make some assumptions without a PR/commit message (hint hint)

cmd/vfkit/main.go
lstocchi: yes, updated. It was fixed on that specific path bc of some specific script I was using for testing 👍 
but yes, there is no reason to keep it fixed, updated to a random filename

cmd/vfkit/main.go
```diff
@@ -264,3 +278,97 @@ func startIgnitionProvisionerServer(ignitionReader io.Reader, ignitionSocketPath
...
+// It also register an exit handler to delete the file when vfkit exits
+func createCloudInitISO(metaData, userData []byte) (string, error) {
+	writer, err := iso9660.NewWriter()
+	if err != nil {
+		return "", errors.Wrap(err, fmt.Sprintf("failed to create writer: %s", err))
+	}
+
+	defer writer.Cleanup()
+
+	userDataReader := bytes.NewReader(userData)
```
cfergeau, line 339: In ` generateCloudInitImage`, you get an `io.Reader` from `os.Open` which is used to read the file content into a `[]byte` variable. Then you pass this variable to `createCloudInitISO` and you create an `io.Reader` from the `[]byte` argument.
I'd change the parameters to `createCloudInitISO` to be `io.Reader`, this should remove some code.

cmd/vfkit/main.go
```diff
@@ -264,3 +278,92 @@ func startIgnitionProvisionerServer(ignitionReader io.Reader, ignitionSocketPath
...
+
+	tmpDir := os.TempDir()
+	isoFile, err := os.CreateTemp(tmpDir, "vfkit-cloudinit-")
+	if err != nil {
+		return "", errors.Wrap(err, "unable to create temporary cloud-init ISO file")
+	}
+
+	outputFile, err := os.OpenFile(isoFile.Name(), os.O_WRONLY|os.O_TRUNC|os.O_CREATE, 0644)
+	if err != nil {
+		return "", errors.Wrap(err, fmt.Sprintf("failed to open cloud-init ISO file: %s", err))
```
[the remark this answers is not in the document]
cfergeau: github.com/pkg/errors has been archived for a few years, better not to start using it.

cmd/vfkit/main.go
```diff
@@ -264,3 +278,97 @@ func startIgnitionProvisionerServer(ignitionReader io.Reader, ignitionSocketPath
...
+	}
+
+	err = writer.WriteTo(outputFile, "cidata")
+	if err != nil {
+		return "", errors.Wrap(err, fmt.Sprintf("failed to write cloud-init ISO image: %s", err))
+	}
+
+	err = outputFile.Close()
```
cfergeau, line 367: should be a `defer`, otherwise we'll leak the file in error cases.

cmd/vfkit/main.go
```diff
@@ -264,3 +278,97 @@ func startIgnitionProvisionerServer(ignitionReader io.Reader, ignitionSocketPath
...
+	if err != nil {
+		return "", errors.Wrap(err, fmt.Sprintf("failed to add meta-data file: %s", err))
+	}
+
+	tmpDir := os.TempDir()
+	isoFile, err := os.CreateTemp(tmpDir, "vfkit-cloudinit-")
+	if err != nil {
+		return "", errors.Wrap(err, "unable to create temporary cloud-init ISO file")
+	}
+
```
cfergeau, line 356: I'd move the call to `util.RegisterExitHandler` here to ensure the file is also deleted when an error occurs.

cmd/vfkit/main.go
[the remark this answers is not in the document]
lstocchi: done 👍

cmd/vfkit/main.go
```diff
@@ -264,3 +276,94 @@ func startIgnitionProvisionerServer(ignitionReader io.Reader, ignitionSocketPath
...
+		if path == "" {
+			continue
+		}
+		file, err := os.Open(path)
+		if err != nil {
+			return "", err
+		}
+		defer file.Close()
+
+		// based on the file name we save its content on the corresponding var
```
cfergeau, line 301: `we save its content ` this part of the comment is no longer accurate.

cmd/vfkit/main.go
```diff
@@ -264,3 +276,94 @@ func startIgnitionProvisionerServer(ignitionReader io.Reader, ignitionSocketPath
...
+		switch filename {
+		case "user-data":
+			userDataReader = file
+		case "meta-data":
+			metaDataReader = file
+		}
+	}
+
+	if userDataReader == nil && metaDataReader == nil {
+		return "", fmt.Errorf("cloud-init needs user-data and meta-data files to work")
```
cfergeau, line 312: I'm a bit confused by the error message and the test. If I'm not mistaken, the test implies that we need user-data **or** meta-data to be provided, but the way I understand the error message,  we expect to get both?

cmd/vfkit/main.go
```diff
@@ -264,3 +276,94 @@ func startIgnitionProvisionerServer(ignitionReader io.Reader, ignitionSocketPath
...
+	if err != nil {
+		return "", fmt.Errorf("unable to create temporary cloud-init ISO file: %w", err)
+	}
+
+	// register handler to remove isoFile when exiting
+	util.RegisterExitHandler(func() {
+		os.Remove(isoFile.Name())
+	})
+
+	outputFile, err := os.OpenFile(isoFile.Name(), os.O_WRONLY|os.O_TRUNC|os.O_CREATE, 0644)
```
cfergeau, line 352: I'm confused by this code, `os.CreateTemp` returns an `os.File` which we can use to write to the file, do we really need to reopen this file?

cmd/vfkit/main_test.go
```diff
@@ -42,3 +44,66 @@ func TestStartIgnitionProvisionerServer(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, ignitionData, body)
 }
+
+func TestGenerateCloudInitImage(t *testing.T) {
+	_, filename, _, ok := runtime.Caller(0)
+	assert.True(t, ok)
+
+	currentDir := filepath.Dir(filename)
+	projectRoot := filepath.Join(currentDir, "../../")
```
cfergeau, line 53: can this be moved to a `getTestAssetsDir` as it's reused by all the tests? An alternative could be to add a `testdata` directory to `cmd/vfkit` which should be easier to locate?
Did you try using https://pkg.go.dev/os#Getwd instead of `runtime.Caller`?

cmd/vfkit/main.go
lstocchi: Cloud-init requires both files to work. They can be empty but they must exist otherwise it won't work.

However I think it's reasonable to allow the user to only pass on file. Maybe you just need to push some customization using the user-data (or the meta-data) and leave the other file empty. This way if we find out that you only have the user-data, we create an empty meta-data file (or vice versa).

On the other hand, if we do not find any of those files, there is most probably an error and the user needs to be informed

cmd/vfkit/main_test.go
lstocchi: i was not sure if adding a folder in cmd/vfkit was the right way to do it in golang, this is why I kept the assets in the test folder. I applied your suggestions but if you prefer to move the files I can do that

cfergeau: I'm also not sure about the recommended way of doing this, let's keep it as it is for now, we can move this around if we are more knowledgeable one day :)

go.mod
```diff
@@ -53,7 +55,6 @@ require (
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/pelletier/go-toml/v2 v2.2.2 // indirect
-	github.com/pkg/errors v0.9.1 // indirect
```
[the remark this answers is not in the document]
lstocchi: done 👍

cfergeau (approved):

cmd/vfkit/main.go
baude: should the input be a dir that contains files and we check for meta and user data as a minimum and proceed? or is it common to have multiples of those? (when I was a cloud-init expert in the day, it was just mostly those two)

lstocchi: i am using this stuff in another tool and I thought the same thing. 
Would it be acceptable to allow the user to pass a folder or files? Or would it be confusing?
`--cloud-init <folder-path>`
and also
`--cloud-init <folder-path>/userdata,<other-folder>/meta-data`

PR #276, opened 2025-03-09
test: Wait for ignition socket before requesting ignition

booxter:
The test sometimes fails with the following error:

Get "http://unix//virtiovsock": dial unix virtiovsock: connect: no such
file or directory

This happens because of a race between the server goroutine and the test
client.

Spotted-At: https://github.com/NixOS/nixpkgs/issues/373493

Signed-off-by: Ihar Hrachyshka <ihar.hrachyshka@gmail.com>

### DIFF
--- a/cmd/vfkit/main_test.go
+++ b/cmd/vfkit/main_test.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -25,6 +26,14 @@ func TestStartIgnitionProvisionerServer(t *testing.T) {
 		err := startIgnitionProvisionerServer(ignitionReader, socketPath)
 		require.NoError(t, err)
 	}()
+
+	// Wait for the socket file to be created before serving, up to 2 seconds
+	for i := 0; i < 10; i++ {
+		if _, err := os.Stat(socketPath); err == nil {
+			break
+		}
+		time.Sleep(200 * time.Millisecond)
+	}
 
 	// Make a request to the server
 	client := http.Client{


cfergeau (commented):
Looks good to me, though it looks like in theory, vfkit itself could hit the same race, but avoiding it would mean reworking the `startIgnitionProvisionerServer` API

booxter:
@cfergeau I know little about the code base, so I may be wrong, but I think the start function itself doesn't have a race - it correctly creates the socket with `Listen`, then start serving handler. What happens with the test failure is that the goroutine that runs the test case, due to parallelism, reaches the line where it calls to HTTP via the socket before the goroutine that starts the server has a change to `Listen` (which would create the socket file).

cfergeau:
> @cfergeau I know little about the code base, so I may be wrong, but I think the start function itself doesn't have a race - it correctly creates the socket with `Listen`, then starts serving handler. What happens with the test failure is that the goroutine that runs the test case, due to parallelism, reaches the line where it calls to HTTP via the socket before the goroutine that starts the server has a chance to `Listen` (which would create the socket file).

To be honest, this will likely never happen, but since `startIgnitionProvisionerServer` is called in a different go routine from the one `vm. Start()` is called in, in theory the vm could be started and try to use ignition before the ignition server is even setup. Since starting the VM takes several seconds, I don't think this will ever happen.

```
diff --git a/cmd/vfkit/main.go b/cmd/vfkit/main.go
index 0cff38d..1e62689 100644
--- a/cmd/vfkit/main.go
+++ b/cmd/vfkit/main.go
@@ -156,6 +156,7 @@ func runVirtualMachine(vmConfig *config.VirtualMachine, vm *vf.VirtualMachine) e
                        }
                        defer file.Close()
                        reader := file
+                       time.Sleep(15 * time.Second)
                        if err := startIgnitionProvisionerServer(reader, vmConfig.Ignition.SocketPath); err != nil {
                                log.Error(err)
                        }
```
might allow to artificially trigger what I'm describing, which is similar to the issue in the test case

booxter:
Ah true if you mean that `vm.Start()` - the consumer of the `startIngition*Server` function - is racy.

PR #283, opened 2025-03-13
Call exit handlers when process exits normally

### LINKED ISSUE #248: The vfkit client socket is not deleted when the VM terminates from within the guest
```
% vfkit --version
vfkit version: v0.6.0
```

To reproduce:

1. Start vfkit with `--virtio-net,unixSocketPath=/path/vmnet.sock`
2. It creates the client socket at /path/vfkit-190b-e485.sock
3. Login to the VM via serial console and run:
   ```
   sudo poweroff
   ```
3. vfkit terminates:
   ```
   $ sudo poweroff
   ubuntu@example-vm:~$ INFO[0028] VM is stopped                                 
   ```

Socket is not removed.

When interrupting vfkit, the socket is removed:

```
example-vm login: INFO[0014] captured interrupt, calling exit handlers and exiting.. 
```

This does not affect the functionality, but since the socket includes a pid, we get new socket per each vfkit run.

The code includes this 5 months old comment:
```
// FIXME: need to remove localSocketPath at process exit
```

evidolob:
Current implementation in exithandler.go is track only signals(SIGTERM and SIGINT), but vfkit process also could exit normally, when VM is turned off by guest OS. This commit add 'defer util.ExecuteExitHandlers()' call in to root command to call exit handlers when process ending.

Fixes #248

### DIFF
--- a/pkg/util/exithandler.go
+++ b/pkg/util/exithandler.go
@@ -41,14 +41,20 @@ func setupExitSignalHandling(doExit bool) {
 	go func() {
 		for sig := range sigChan {
 			log.Printf("captured %v, calling exit handlers and exiting..", sig)
-			exitRegistry.mutex.Lock()
-			for _, handler := range exitRegistry.handlers {
-				handler()
-			}
-			exitRegistry.mutex.Unlock()
+			ExecuteExitHandlers()
 			if doExit {
 				os.Exit(1)
 			}
 		}
 	}()
+}
+
+// ExecuteExitHandlers is call all registered exit handlers
+// This function should be called when program finish work(i.e. when VM is turned off by guest OS)
+func ExecuteExitHandlers() {
+	exitRegistry.mutex.Lock()
+	for _, handler := range exitRegistry.handlers {
+		handler()
+	}
+	exitRegistry.mutex.Unlock()
 }
--- a/cmd/vfkit/root.go
+++ b/cmd/vfkit/root.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/crc-org/vfkit/pkg/cmdline"
+	"github.com/crc-org/vfkit/pkg/util"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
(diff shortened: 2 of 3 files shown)


lstocchi (commented):
/lgtm

pkg/util/exithandler.go
```diff
@@ -41,14 +41,20 @@ func setupExitSignalHandling(doExit bool) {
 	go func() {
 		for sig := range sigChan {
 			log.Printf("captured %v, calling exit handlers and exiting..", sig)
-			exitRegistry.mutex.Lock()
-			for _, handler := range exitRegistry.handlers {
-				handler()
-			}
-			exitRegistry.mutex.Unlock()
+			ExecuteExitHandlers()
 			if doExit {
```
nirs, line 45: When doExit is false, the exit handlers will run twice. It would be better to run them exactly once, either here, or in main().

Since the handlers are running under a mutex, we can clear the exitRegistry ensuring that handlers run exactly once.

nirs: But I think we have a better way to handle this. The root cause is that we don't do graceful shutdown when receiving a signal, which is the expected behaviour. #284.

We can change the signal handler to run graceful shutdown (same as POST Stop). Then perform all exit handlers in main when runVfkit() returns. This will solve both issue with simpler code.

cfergeau: > We can change the signal handler to run graceful shutdown (same as POST Stop). Then perform all exit handlers in main when runVfkit() returns. 

While I generally agree with what you wrote here and in #284 , there’s an additional complication with graceful shutdown: the guest may not shutdown (eg if there was a kernel panic). The corresponding API is https://developer.apple.com/documentation/virtualization/vzvirtualmachine/requeststop()?language=objc , the terminology it uses is to "request", "ask" the guest OS to stop.

If `runVfkit`  does not return after this request has been sent, we’ll have to decide if we force kill the VM or if we let that up to the user, in which case the cleanups won’t be done.

cfergeau: > When doExit is false, the exit handlers will run twice. It would be better to run them exactly once, either here, or in main().
> 
> Since the handlers are running under a mutex, we can clear the exitRegistry ensuring that handlers run exactly once.

`doExit` is only false in the testing code, since I’d like to make a vfkit release, I’ll merge this as is, and this can be reworked as part of https://github.com/crc-org/vfkit/issues/284

nirs: Sure, Stop may not happen, and this matches the semantics of SIGTERM, the process may handle it and never exit. This is typically enforced by SIGKILL after a timeout. So internally vfkit should probably use a timeout and fallback to HardStop, since we don't want to be killed without being able to clean up.

Issue #242, opened 2025-01-12
macOS 15 doesn't set mac addr in dhcp_leases anymore


phaer:
Hello,

`usage.md` recommends `/var/db/dhcp_leases` to get the IP address of a given virtual machine.
But it seems like macOS Sequoia has changed the format of that file and haven't been able to find much about it nor to easily match DHCP leases with VMs.

```
{
        name=nixos
        ip_address=192.168.64.30
        hw_address=ff,f1:f5:dd:7f:0:2:0:0:ab:11:b5:40:7c:3:e2:a9:8c:b3
        identifier=ff,f1:f5:dd:7f:0:2:0:0:ab:11:b5:40:7c:3:e2:a9:8c:b3
        lease=0x6782bfd4
}
{
        name=nixos
        ip_address=192.168.64.29
        hw_address=1,f6:25:e2:48:58:1e
        identifier=1,f6:25:e2:48:58:1e
        lease=0x6782bf9a
}
```

The second entry is an older lease, probably from before I upgraded? It contains the mac address of my machine as expected. The first entry is a current lease, without a mac address. At least not a plain one, I haven't been able to find documentation about the new format.

https://github.com/canonical/multipass/issues/3661 looks like it could be related.
I wonder if might be a side-effect of https://macadmin.fraserhess.com/2024/09/16/handling-mac-address-randomization-in-macos-15/.

lstocchi:
Hi phaer,

thanks for opening the issue. It could be a side-effect of the MAC address randomization as you pointed out. We'll work on an enhancement to help users find the VM IP. 

Were you still able to log in into the VM using the first entry, though?

phaer:
> Were you still able to log in into the VM using the first entry, though?

Yes, the VM is reachable just fine. 

A workaround to get a predictable IP for the VM is to use a static mac address and convert that to a ipv6 link local ip (e.g. via ipv6calc or one the online calculators).

gbraad:
Looks to work like a stack (populates at the top).
And the lease looks like a hex (timestamp?)

`0x6782bfd4` = 1736622036 => _Sat Jan 11 2025 19:00:36 GMT+0000_

gbraad:
> The second entry is an older lease, probably from before I upgraded?

Unlikely, as it is a timestamp from just a minute before. @phaer Was this a single start of the VM ? As I am confused why you got two different entries. Can you explain more how you were connected?

cmoulliard:
[...]

PR #286, opened 2025-03-24
Add a script to start vfkit with cloud init configuration

vyasgun:
This script does the following:

- Accept values of ssh key, ssh user, disk image path from the user
- Generate `user-data` and `meta-data` files
- Create an image and mount it as a block device
- Start vfkit VM with the above configuration


lstocchi (commented):
/lgtm

contrib/scripts/start-with-cloud-init.sh
```diff
@@ -0,0 +1,58 @@
...
+# $ SSH_USER=test HOST_NAME=vm1 DISK_IMG=Fedora-Cloud-Base-AmazonEC2-41-1.4.aarch64.raw \
+# SSH_PUB_KEY=id_rsa.pub ./contrib/scripts/start-with-cloud-init.sh
+#
+# $ ssh -i id_rsa test@192.168.64.14
+
+set -exu
+
+HOST_NAME=${HOST_NAME:-"vfkit-vm"}
+
+PUBLIC_KEY=$(cat "$SSH_PUB_KEY")
```
cfergeau, line 28: Maybe you could test that `SSH_PUB_KEY` is set before attempting to read it, and error out if not?

contrib/scripts/start-with-cloud-init.sh
```diff
@@ -0,0 +1,58 @@
...
+
+mkdir -p out
+
+cat <<EOF > out/user-data
+#cloud-config
+users:
+  - name: $SSH_USER
```
cfergeau, line 35: I’d add a default value for `SSH_USER`, or make sure to error out if it is unset.

contrib/scripts/start-with-cloud-init.sh
```diff
@@ -0,0 +1,58 @@
...
+  expire: false
+EOF
+
+cat <<EOF > out/meta-data
+instance-id: $HOST_NAME
+local-hostname: $HOST_NAME
+EOF
+
+./out/vfkit --cpus 2 --memory 2048 \
+    --bootloader efi,variable-store=efistore.nvram,create \
```
cfergeau, line 53: `efistore.nvram` can also be created in `out/`

cfergeau (commented):
Coincidentally, https://github.com/nirs/vmnet-helper/pull/63/files contains a similar script (no big impact for this PR, just 'fyi'

contrib/scripts/start-with-cloud-init.sh
```diff
@@ -0,0 +1,58 @@
...
+    --cloud-init out/user-data,out/meta-data \
+    --device virtio-blk,path="$DISK_IMG" \
```
cfergeau, line 55: you could also check that `DISK_IMG` is set and exists.

contrib/scripts/start-with-cloud-init.sh
```diff
@@ -0,0 +1,70 @@
...
+cat <<EOF > out/user-data
+#cloud-config
+users:
+  - name: $SSH_USER
+    sudo: ALL=(ALL) NOPASSWD:ALL
+    groups: users
+    shell: /bin/bash
+    lock_passwd: true
+    ssh_authorized_keys:
+      - "$PUBLIC_KEY"
```
nirs, line 53: You don't need to create new user, the minimal example for cloud-init is:

```yaml
#cloud-config
password: password
chpasswd:
  expire: false
ssh_authorized_keys:
- "$PUBLIC_KEY"
```

contrib/scripts/start-with-cloud-init.sh
```diff
@@ -0,0 +1,70 @@
...
+ssh_pwauth: false
+chpasswd:
+  expire: false
+EOF
+
+cat <<EOF > out/meta-data
+instance-id: $HOST_NAME
+local-hostname: $HOST_NAME
+EOF
```
nirs, line 62: We want at least this network-config:

```bash
cat > network-config << EOF
version: 2
ethernets:
  eth0:
    match:
      macaddress: $MAC_ADDRESS
    set-name: eth0
    dhcp4: true
    dhcp-identifier: mac
EOF
```

The "dhcp-indetifier: mac" fixes #242.

contrib/scripts/start-with-cloud-init.sh
```diff
@@ -0,0 +1,70 @@
...
+
+./out/vfkit --cpus 2 --memory 2048 \
+    --bootloader efi,variable-store=out/efistore.nvram,create \
+    --cloud-init out/user-data,out/meta-data \
```
nirs, line 66: This should allow also network-config.

contrib/scripts/start-with-cloud-init.sh
cfergeau: But is the default user name for images using cloud-init standard across distros? imo it’s simpler if we specify it in this script, this way we  won’t have to try `cloud-init`, `cloudinit`, `cloud`, `root`, `ubuntu`, … until we find one which works.

contrib/scripts/start-with-cloud-init.sh
cfergeau: I’ve created https://github.com/crc-org/vfkit/issues/296 for this. I don’t think we need to hold this PR for this, we can update the example script when the support is there.

cfergeau (approved):

cfergeau:
/hold
waiting a bit to (auto) merge this until this conversation is resolved https://github.com/crc-org/vfkit/pull/286#discussion_r2034131789

cfergeau:
Ah, I was not fast enough with the /hold :) it’s merged already
/unhold

Issue #275, opened 2025-03-06
Add script to boot a VM with ignition and a user-provided (or auto-generated) SSH key


cfergeau:
This would make testing of coreos images easier (and maybe other kind of images)

cfergeau:
something like `SSH_KEY=~/.ssh/id_rsa SSH_USER=core vfkit-cloud-init.sh rhel-cloud.qcow2` (just an example, the names can be different, can be cmdline args instead of env vars, …)

PR #287, opened 2025-03-24
Add script to boot VM with ignition and user provided ssh key

vyasgun:
Fixes: https://github.com/crc-org/vfkit/issues/275
The script does the following:
- Accept SSH user and key from the user as environment variables
- Generate an ignition file with the values set
- Start vfkit VM with the above configuration

### DIFF
--- a/contrib/scripts/start-with-ignition.sh
+++ b/contrib/scripts/start-with-ignition.sh
@@ -1,0 +1,65 @@
+#!/bin/sh
+
+# This script can be used to start a vfkit VM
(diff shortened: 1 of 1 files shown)


lstocchi (commented):
LGTM
I would add some comment in the header like the `start_crc_bundle.sh`

vyasgun:
Thanks @lstocchi - updated the script with comment

lstocchi (commented):
/lgtm

cfergeau:
I have the same comments as for the cloud-init PR.

PR #292, opened 2025-03-31
pkg/rest: refactor and simplify

kolyshkin:
This removes unused code from pkg/rest, and simplifies it.

Should not change functionality in any way.

Please see individual commits for details.

## Summary by Sourcery

Refactor and simplify the REST package by removing unused code and consolidating URI parsing logic

Enhancements:
- Simplified URI parsing logic by merging validation steps into the NewEndpoint function
- Removed case-insensitive scheme parsing and consolidated scheme conversion

Tests:
- Updated test cases to reflect new Endpoint creation and scheme parsing logic

Chores:
- Removed unused functions like validateRestfulURI and parseRestfulURI
- Removed unused import for cmdline package


pkg/rest/rest.go
```diff
@@ -112,55 +132,15 @@ type VirtualMachineStateHandler interface {
...
-// toRestScheme converts a string to a ServiceScheme
+// toRestScheme converts a string to a [ServiceScheme].
 func toRestScheme(s string) (ServiceScheme, error) {
-	switch strings.ToUpper(s) {
-	case "NONE":
+	switch s {
+	case "none":
 		return None, nil
-	case "UNIX":
+	case "unix":
```
gbraad, line 140: Can the scheme be used?

pkg/rest/rest.go
```diff
@@ -112,55 +132,15 @@ type VirtualMachineStateHandler interface {
...
-	if scheme == Unix && len(restURI.Path) > maxSocketPathLen() {
-		return nil, fmt.Errorf("invalid unix uri: socket path length exceeds macOS limits")
-	}
-	return restURI, err
-}
-
-// toRestScheme converts a string to a ServiceScheme
+// toRestScheme converts a string to a [ServiceScheme].
 func toRestScheme(s string) (ServiceScheme, error) {
-	switch strings.ToUpper(s) {
```
gbraad, line 148: This is to ensure no mistakes are made when this is specified manually.
Either use `upper` or `lower`

pkg/rest/rest.go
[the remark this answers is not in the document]
kolyshkin: Do you mean, a const from the standard packages? I am not aware of any, please suggest.

Or do you mean to introduce and use own constant? If that's the case, I don't think it makes sense here.

pkg/rest/rest.go
kolyshkin: Can you point me out to a case in the code where it is specified manually?

From what I see, it is only used to parse a Scheme field from a structure obtained from url.ParseRequestURI, which ensures Scheme is lowercase, so we can expect the lowercased string here. This was noted in the commit message, too.

gbraad: I'll let @cfergeau decide as he maintains this codebase. 

I just notice that https://github.com/crc-org/vfkit/pull/292/files#diff-e8eb5f08837c85d4c2db17fa463051c4a787efb7cbc9883092f53b24a48514e1R32 uses the input that comes from the request parse; and it assumes the format to be returned always to be lower. Defensive programming is not wrong, as this is not code to be performant or such (and mistakes have been made before). Overall, this looks more maintainable.

pkg/rest/rest.go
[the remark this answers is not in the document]
gbraad: I thought they had... https://pkg.go.dev/net/url#pkg-constants "Section is emptry" :-/

gbraad (approved):
Haven't tested functionality, but looks cleaner.

pkg/rest/rest.go
```diff
@@ -25,14 +25,36 @@ type Endpoint struct {
...
 	if err != nil {
 		return nil, err
 	}
 	scheme, err := toRestScheme(uri.Scheme)
 	if err != nil {
 		return nil, err
 	}
+	switch scheme {
+	case TCP:
+		if len(uri.Host) < 1 {
```
cfergeau, line 38: `toRestScheme` is only called once after this change, and it is mostly checking that the uri scheme is a valid one, and then in this `switch`/`case`, we do more validation on the URI.
I’d move `toRestScheme` and the validation code from `parseRestfulURI` to a single `validateURI(uri)` method which would do everything at once.
I prefer to have the validation code in a separate function to avoid cluttering `NewEndpoint`.

cfergeau (commented):
3rd commit can be improved in my opinion, the first 2 look good to me.

kolyshkin:
> 3rd commit can be improved in my opinion, the first 2 look good to me.

I have remove the third commit.

pkg/rest/rest.go
kolyshkin: I've removed the third commit.

cfergeau:
> > 3rd commit can be improved in my opinion, the first 2 look good to me.
> 
> I have remove the third commit.

Are you planning to do some work on it? Or should we merge this as is?

Issue #278, opened 2025-03-10
Add "pid management" code to vfkit/the client API?


cfergeau:
This came up here https://github.com/kubernetes/minikube/pull/20506#discussion_r1987078966

nirs:
I think this is not only the pid management code, but vfkit can provide a library to run the vfkit process. Many clients need to build the command line arguments, run the process, monitor the state, and stop it. Having the vfkit library will make this easier to integrate correctly.

cfergeau:
For reference, https://github.com/kubernetes/minikube/pull/20501 has some `pidfile` code. crc also has code searching for the vfkit process, ... which should be similar.

nirs:
@cfergeau Please check https://github.com/kubernetes/minikube/pull/20528, it can be reused in vfkit.

PR #293, opened 2025-04-04
Add a Process struct for managing processes by pid file

vyasgun:
This commit does the following:
- Adds Process struct
- Read and write to a pidFile specified by the user
- Adds tests for process lifecycle with a sleep process

Fixes: https://github.com/crc-org/vfkit/issues/278 
Fixes: https://github.com/crc-org/vfkit/issues/309

## Summary by Sourcery

Implement a ProcessManager interface for managing processes using pid files

New Features:
- Create a Process struct and Manager interface to handle process lifecycle management using pid files

Tests:
- Add comprehensive test suite for process management, including process creation, finding, checking existence, and killing processes

## Summary by CodeRabbit

- New Features
  - Added a --pidfile option to save the VM process ID to a specified file, enabling easier monitoring and external lifecycle management.
  - The application now writes its PID to the provided file on startup when the option is used.

- Tests
  - Added comprehensive tests for PID file handling and process lifecycle behaviors to ensure reliability of the new option and related functionality.

### DIFF
--- a/cmd/vfkit/main.go
+++ b/cmd/vfkit/main.go
@@ -35,6 +35,7 @@ import (
 	"github.com/Code-Hex/vz/v3"
 	"github.com/crc-org/vfkit/pkg/cmdline"
 	"github.com/crc-org/vfkit/pkg/config"
+	"github.com/crc-org/vfkit/pkg/process"
 	"github.com/crc-org/vfkit/pkg/rest"
 	restvf "github.com/crc-org/vfkit/pkg/rest/vf"
 	"github.com/crc-org/vfkit/pkg/vf"
@@ -103,6 +104,19 @@ func newVMConfiguration(opts *cmdline.Options) (*config.VirtualMachine, error) {
 		opts.Devices = append(opts.Devices, fmt.Sprintf("virtio-blk,path=%s", cloudInitISO))
 	}
 
+	if opts.PidFile != "" {
+		execPath, err := os.Executable()
+		if err != nil {
+			return nil, fmt.Errorf("could not determine executable path: %w", err)
+		}
+		vfProcess := process.New(os.Args[0], opts.PidFile, execPath)
+		pid := os.Getpid()
+		err = vfProcess.WritePidFile(pid)
+		if err != nil {
+			return nil, fmt.Errorf("could not write PID: %w", err)
+		}
+	}
+
 	if err := vmConfig.AddDevicesFromCmdLine(opts.Devices); err != nil {
 		return nil, err
 	}
@@ -110,7 +124,6 @@ func newVMConfiguration(opts *cmdline.Options) (*config.VirtualMachine, error) {
 	if err := vmConfig.AddIgnitionFileFromCmdLine(opts.IgnitionPath); err != nil {
 		return nil, fmt.Errorf("failed to add ignition file: %w", err)
 	}
-
 	return vmConfig, nil
 }
 
--- a/pkg/cmdline/cmdline.go
+++ b/pkg/cmdline/cmdline.go
@@ -29,6 +29,8 @@ type Options struct {
 	CloudInitFiles stringSliceValue
 
 	Nested bool
+
+	PidFile string
 }
 
 const DefaultRestfulURI = "none://"
@@ -59,4 +61,5 @@ func AddFlags(cmd *cobra.Command, opts *Options) {
 	cmd.Flags().StringVar(&opts.IgnitionPath, "ignition", "", "path to the ignition file")
 	cmd.Flags().VarP(&opts.CloudInitFiles, "cloud-init", "", "path to user-data and meta-data cloud-init configuration files")
 	cmd.Flags().BoolVarP(&opts.Nested, "nested", "n", false, "enable nested virtualization")
+	cmd.Flags().StringVar(&opts.PidFile, "pidfile", "", "path to the pid file")
 }
--- a/pkg/config/json_test.go
+++ b/pkg/config/json_test.go
@@ -231,7 +231,7 @@ var jsonStabilityTests = map[string]jsonStabilityTest{
 
 			return vm
 		},
-		skipFields:   []string{"Bootloader", "Devices", "Timesync", "Ignition", "Nested"},
+		skipFields:   []string{"Bootloader", "Devices", "Timesync", "Ignition", "Nested", "PidFile"},
 		expectedJSON: `{"vcpus":3,"memoryBytes":3,"bootloader":{"kind":"linuxBootloader","vmlinuzPath":"/vmlinuz","kernelCmdLine":"console=hvc0","initrdPath":"/initrd"},"devices":[{"kind":"virtiorng"}],"timesync":{"vsockPort":1234}}`,
 	},
 	"RosettaShare": {


vyasgun:
@sourcery-ai review

pkg/process/process.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (p *Process) Pid() (int, error) {
+	data, err := os.ReadFile(p.PidFilePath())
+	if err != nil {
+		return -1, err
+	}
+	pidStr := strings.TrimSpace(string(data))
+	pid, err := strconv.Atoi(pidStr)
+	return pid, errors.Wrap(err, "invalid pid file")
```
cfergeau, line 46: `errors.Wrap` is deprecated, https://github.com/pkg/errors is archived. This has been superseded by https://go.dev/blog/go1.13-errors

pkg/process/process.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	return proc, nil
+}
+
+func (p *Process) Exists() bool {
+	proc, err := p.FindProcess()
+	if err != nil {
+		return false
+	}
+	if err := proc.Signal(syscall.Signal(0)); err != nil {
```
cfergeau, line 67: Why `syscall.Signal(0)`? Could this use one of the `SIGXXX` constants from https://pkg.go.dev/syscall#pkg-constants ? (search for `SIGABRT`)

cfergeau (commented):
Have you tried using this to replace the equivalent code in crc, or possibly in podman? This would be the best way to make sure the interface is the one we need.

pkg/process/process.go
vyasgun: Is it safe to call SIGABRT on the running process? On searching I found sending 0 is the safest way to check a process is running

pkg/process/process.go
```diff
@@ -0,0 +1,90 @@
...
+		return false
+	}
+	return true
+}
+
+func (p *Process) Kill() error {
```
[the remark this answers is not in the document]
vyasgun: I'll add `Terminate` as well

pkg/process/process.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (p *Process) Name() string {
+	return p.name
+}
+
+func (p *Process) PidFilePath() string {
+	return p.pidFilePath
+}
+
+func (p *Process) Pid() (int, error) {
```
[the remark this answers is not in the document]
vyasgun: Thanks, noted.

pkg/process/process.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	proc, err := os.FindProcess(pid)
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot find process")
+	}
+	return proc, nil
+}
+
+func (p *Process) Exists() bool {
```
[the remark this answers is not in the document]
vyasgun: Okay, I'll take that into account and modify the function

pkg/process/process.go
cfergeau: I mentioned `SIGABRT` as a way to find the relevant section, I was not suggesting to use this signal :)
However, if `FindProcess()` found it, I’d assume it means the process is running, no need to check again that it’s running ?

pkg/process/process_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+func TestProcess_KillProcess(t *testing.T) {
+	err := managedProcess.Kill()
+	assert.NoError(t, err)
+	assert.False(t, managedProcess.Exists())
+
+	// Try to kill the non-existent process
+	// This should result in an error
+	err = managedProcess.Kill()
+	assert.Error(t, err)
```
nirs, line 80: Kill will be easier to use if it is idempotent. Trying to kill a process that does not exist is not an error from the caller point of view.

pkg/process/process_test.go
```diff
@@ -0,0 +1,99 @@
+package process
```
nirs, line 1: Looks like we need only public names, so we can use process_test package. This will be better tests since the test code will be the same as real code using this package.

pkg/process/process.go
```diff
@@ -0,0 +1,119 @@
+package process
```
nirs, line 1: No copyright/license info?

pkg/process/process_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	os.Exit(exitCode)
+}
+
+func TestProcess_Name(t *testing.T) {
+	assert.Equal(t, dummyProcessName, managedProcess.Name())
+}
+
+func TestProcess_FindProcess(t *testing.T) {
```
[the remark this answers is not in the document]
nirs: This suggestion is not good.
- Bad test names
- Not using t.TempDir()
- Not using os.WriteFile
- Closing stuff manually without defer
- Adding unneeded dependencies (testify, require)

pkg/process/process_test.go
```diff
@@ -0,0 +1,99 @@
...
+	assert.NoError(t, err)
+	assert.False(t, managedProcess.Exists())
+
+	// Try to kill the non-existent process
+	// This should result in an error
+	err = managedProcess.Kill()
+	assert.Error(t, err)
+}
+
+func TestProcess_FindProcess_InvalidPidFile(t *testing.T) {
```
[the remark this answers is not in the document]
nirs: More bad advice.

nirs (commented):
We need tests showing the expected usage of this package. You can check minikube vfkit driver to see the use cases that this package should support.

pkg/process/process_test.go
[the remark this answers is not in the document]
nirs: Much better!

But the test name `TestProcess_FindProcess_InvalidPid(t *testing.T)` is too long and and mixing CamelCase and underscore is the worst possible code style.

Try to use sub tests to make this nicer. We want to test FindProcess, and there are many use cases:
- no pid file
- invalid pid file
- stale pid (pid used by another process)
- permission error when reading pidfile

pkg/process/process.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	pidStr := strings.TrimSpace(string(data))
+	pid, err := strconv.Atoi(pidStr)
+	if err != nil {
+		return -1, fmt.Errorf("invalid pid file: %v", err)
+	}
+	return pid, nil
+}
+
+func (p *Process) FindProcess() (*process.Process, error) {
```
nirs, line 58: Does it need to be public or it an be a private helper? Exposing the underlying process package in this packages means we can never change the implementation, for example using another package to validate the process executable path.

pkg/process/process.go
```diff
@@ -0,0 +1,90 @@
...
+	WritePidFile(pid int) error
+}
+
+type Process struct {
+	name        string
+	pidFilePath string
+}
+
+func New(name, pidFilePath string) (*Process, error) {
+	return &Process{name: name, pidFilePath: pidFilePath}, nil
```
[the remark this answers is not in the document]
nirs: Good advice! the error is not needed, no reason to add it.

pkg/process/process_test.go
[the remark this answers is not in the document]
nirs: It will be easier to use if New was creating the struct without any checks, and Process.Exist() would do the checks.

For example lets say we check in New() that pidfile and exectablePath exists, but after we check they are deleted. How the check helped the user? It will be much easier if we check that should be check when calling Exists().

Testing missing or inaccessible pidfile and executable path sounds good.

pkg/process/process.go
```diff
@@ -0,0 +1,90 @@
...
+	return p.pidFilePath
+}
+
+func (p *Process) Pid() (int, error) {
+	data, err := os.ReadFile(p.PidFilePath())
+	if err != nil {
+		return -1, err
+	}
+	pidStr := strings.TrimSpace(string(data))
+	pid, err := strconv.Atoi(pidStr)
```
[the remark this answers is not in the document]
nirs: It may help - need to test invalid pid with real code using the library and see if the error makes sense or we need to add some context.

pkg/process/process_test.go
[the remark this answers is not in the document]
nirs: This table test is the worst way - it is very hard to understand the complicated test cases with so much info in each case, name, setup, booleans. And the code it much more complicated than needed since we must support all cases in the table.

This will be easier to follow and work with with nested tests using simpler code and having the setup code near the assertions.

pkg/process/process.go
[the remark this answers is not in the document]
nirs: Why add an old dependency (errors.Wrapf) when we can use standard library (fmt.Errorf("...: %w", err))?

pkg/process/process_test.go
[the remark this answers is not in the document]
nirs: This looks like the right direction.

This setup code is repeated in all tests:

```
dir := t.TempDir()
pidFile := filepath.Join(dir, "locked.pid")
if err := os.WriteFile(pidFile, []byte(strconv.Itoa(os.Getpid())), 0o000); err != nil {
    t.Fatalf("WriteFile(): %v", err)
}
```

It is not very interesting, but we need it for all tests. We can extract this to a helper creating a pidfile in the test temporary directory, so for most tests we will have single line for creating the pidfile:

    pidfile := writePidfile(t, "42", 0o600)

With this we can create all cases:
- invalid pid file
- missing pid file
- inaccessible pid file
- valid pid file
- stale pid file

If writePidfile fails, it can fail the test.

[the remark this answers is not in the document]
nirs: @vyasgun please check

pkg/process/process.go
vyasgun: In crc, there's a codepath that uses something like `FindProcess` to get the underlying Process object:
https://github.com/crc-org/crc/compare/main...vyasgun:crc:test-process-lib

I think this can be changed in crc so the code doesn't use the underlying `process.Process` object if it's not recommended to return it.

pkg/process/process_test.go
```diff
@@ -0,0 +1,166 @@
...
+	cmd := exec.Command("go", "build", "-o", out, source)
+	cmd.Env = append(os.Environ(),
+		"GOOS="+runtime.GOOS,
+		"GOARCH="+runtime.GOARCH,
+	)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	if err := cmd.Run(); err != nil {
+		return "", err
```
nirs: This helper will be more useful if it fails the test instead of returning an error. For example:

    t.Fatal(err)

test/assets/sleep/sleeper.go
nirs, line 1: Maybe put this in process/testdata/sleep.go?

pkg/process/process_test.go
```diff
@@ -0,0 +1,166 @@
...
+		exists, err := process.Exists()
+		assert.NoError(t, err)
+		assert.False(t, exists)
+	})
+}
+
+func build(t *testing.T, name string) (string, error) {
+	source := getSource(t, name)
+
+	out := filepath.Join(source, name)
```
nirs: Using the t.TempDir() will make this easier to use in the tests.

pkg/process/process_test.go
```diff
@@ -0,0 +1,166 @@
...
+	if runtime.GOOS == "windows" {
+		out += ".exe"
```
nirs: vfkit runs on Windows?

pkg/process/process_test.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	t.Logf("Building %q", name)
+	cmd := exec.Command("go", "build", "-o", out, source)
+	cmd.Env = append(os.Environ(),
+		"GOOS="+runtime.GOOS,
+		"GOARCH="+runtime.GOARCH,
+	)
```
nirs: Why do we need this? vfkit is macOS only, and go build does the right thing without changing the env.

pkg/process/process_test.go
```diff
@@ -0,0 +1,166 @@
...
+		assert.NoError(t, err)
+		waitForTermination(t, cmd)
+		exists, err := process.Exists()
+		assert.NoError(t, err)
+		assert.False(t, exists)
+	})
+	t.Run("kill", func(t *testing.T) {
+		cmd := startProcess(t, sleep)
+		pid := cmd.Process.Pid
+		if pid < 0 || pid > math.MaxInt32 {
```
nirs, line 78: pid cannot be 0, but this check is not needed

pkg/process/process_test.go
```diff
@@ -0,0 +1,166 @@
...
+	assert.Error(t, err)
+	assert.ErrorContains(t, err, "invalid pid file")
+}
+
+func TestProcess(t *testing.T) {
+	sleep, err := build(t, "sleep")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(sleep)
```
nirs: Can be simplified by building sleep inside t.TempDir() that will be removed automatically when the test ends.

pkg/process/process_test.go
```diff
@@ -0,0 +1,166 @@
...
+	t.Logf("Started process %q (pid=%v)", name, cmd.Process.Pid)
+
+	t.Cleanup(func() {
+		_ = cmd.Process.Kill()
+		_ = cmd.Wait()
+	})
+	return cmd
+}
+
+func getTestAssetsDir() (string, error) {
```
nirs: The common location is "testdata", under the package.

pkg/process/process.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	if proc == nil {
+		return nil, os.ErrNotExist
+	}
+	name, err := proc.Name()
+	if err != nil {
+		return nil, fmt.Errorf("cannot find process name: %v", err)
+	}
+	if name != p.Name {
+		return nil, fmt.Errorf("process name mismatch. pid %d is being used by %s", pid, name)
```
cfergeau: My feeling is that this should be an `os.ErrNotExist` or such, the process with `pid` is not the one we expect, this means "our" process most likely stopped at some point.

pkg/process/process.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	exe, err := proc.Exe()
+	if err != nil {
+		return nil, fmt.Errorf("cannot find process exe: %v", err)
+	}
+	if exe != p.ExecutablePath {
+		return nil, fmt.Errorf("executable mismatch. pid %d is being used by %s", pid, exe)
```
cfergeau: Same comment as earlier.

pkg/process/process_test.go
cfergeau: It’s needed to avoid gosec overflow warnings when casting `pid` to `int32`. However, this could be an indication that `WritePidFile` should accept an `int`, and do the cast in the method if it’s needed.

pkg/process/process_test.go
cfergeau: vfkit only runs on macos, but could be useful if we reuse this code elsewhere.

test/assets/sleep/sleeper.go
```diff
@@ -0,0 +1,9 @@
+package main
+
+import (
+	"time"
+)
+
+func main() {
+	time.Sleep(1 * time.Minute)
```
cfergeau: Wondering if we need this helper, or if we should be calling `sleep` directly on macos/linux, and `timeout` on windows

pkg/process/process_test.go
nirs: But this is test code that nobody else will use.

pkg/process/process_test.go
nirs: Do we need to cast to int32? We do't have this issue in minikube:
https://github.com/kubernetes/minikube/blob/64e40c6190c65dda679c908414e2de10035f25d7/pkg/minikube/process/process.go#L31

We write pid value from os.Getpid() to disk and read it back. No validation is needed.

vyasgun: I changed it to int. I was using int32 because we are using `process.Process` which expects int32 but we can have the conversion in just one place when it's needed.

pkg/config/config.go
```diff
@@ -33,6 +33,7 @@ type VirtualMachine struct {
 	Timesync   *TimeSync      `json:"timesync,omitempty"`
 	Ignition   *Ignition      `json:"ignition,omitempty"`
 	Nested     bool           `json:"nested,omitempty"`
+	PidFile    string         `json:"pidFile,omitempty"`
```
cfergeau: This field never seem to be used?

pkg/process/process.go
```diff
@@ -0,0 +1,120 @@
...
+func (p *Process) findProcess() (*process.Process, error) {
+	pid, err := p.ReadPidFile()
+	if err != nil {
+		return nil, err
+	}
+	if pid < 0 || pid > math.MaxInt32 {
+		return nil, fmt.Errorf("invalid pid: %d", pid)
+	}
+	proc, err := process.NewProcess(int32(pid))
+	if err != nil && err.Error() == "process does not exist" {
```
cfergeau, line 62: This can be `error.Is(err, process.ErrProcessNotRunning)`

cfergeau (commented):
- Some of the changes in the second commit seem to belong in the first (file rename, test changes, …)
- The suggestion from coderabbit to "Consider using standard library error wrapping." is worth following (use `%w` instead of `%v` when returning errors with `fmt.Errorf`)

cfergeau:
First commit also needs `go mod tidy` to be run

Issue #290, opened 2025-03-31
Error out when a qcow2 image is detected


cfergeau:
Trying to start vfkit with a qcow2 image is an easy mistake to make. We could have detection code to error out early when we get such an image. Maybe https://github.com/lima-vm/go-qcow2reader has an API for this.

nirs:
The [image package](https://github.com/lima-vm/go-qcow2reader/tree/master/image) does what we need. All formats expect qcow2 are stubs that implement only simple probing to be able to detect a raw image.

We can create a probe package by removing the code for reading qcow2 images. We can also make this reusable in go-qcow2reader but I'm not sure it make sense there to extract probing from the actual image drivers.

gbraad:
If more checks are needed, a library makes sense. Otherwise it is overkill. The OP was specific only to check for qcow. @cfergeau  WDYS?

nirs:
> If more checks are needed, a library makes sense. Otherwise it is overkill. The OP was specific only to check for qcow. [@cfergeau](https://github.com/cfergeau) WDYS?

I reported the issue about qcow2 image (qcow is old version which not relevant these days).

I think qcow2 is more important to detect since this is a common image format for appliances. For example cloud images are provided in raw or qcow2 format.

There are other popular virtualization solutions (vmware, parallels, ...) but I don't think users of these solutions will try to use vfkit directly. They are more likely to use some management program (e.g. UTM, lima, minikube) that will handle image format.

vyasgun:
I have created a PR to check for `qcow2` images. I don't think we need to use https://github.com/lima-vm/go-qcow2reader as it seems to be used more for extracting data from the image while we only need the header.

PR #300, opened 2025-04-25
feat: error out on qcow2 block devices

vyasgun:
- Error out when qcow2 image is detected in block device validation
- Fixes: https://github.com/crc-org/vfkit/issues/290


pkg/config/config_test.go
```diff
@@ -44,3 +45,28 @@ func TestNetworkBlockDevice_NoDevice(t *testing.T) {
...
+
+	dev, err := VirtioBlkNew(file.Name())
+	require.NoError(t, err)
+	vm.Devices = append(vm.Devices, dev)
+
+	require.Error(t, vm.ValidBlkDevices())
+
+	// overwrite with non-qcow header
+	file.Seek(0, 0)
+	_, err = file.Write([]byte("NotQcow2"))
```
nirs, line 68: raw images first 4 bytes are zeroes.

pkg/config/config_test.go
```diff
@@ -44,3 +45,28 @@ func TestNetworkBlockDevice_NoDevice(t *testing.T) {
...
+
+func TestVirtualMachine_ValidBlkDevices(t *testing.T) {
+	vm := &VirtualMachine{}
+
+	file, err := os.CreateTemp("", "test-*.img")
+	require.NoError(t, err)
+	defer os.Remove(file.Name())
+
+	// write qcow2 magic number
+	_, err = file.Write([]byte("QFI\xfb"))
```
nirs, line 57: This is not a good way to test. We duplicate the knowledge in the test, if our assumption is incorrect or incomplete the test will pass when the actual code is failing to detect the image.

The best way to ensure that we detect qcow2 images is to create a qcow2 image with tool like qemu-img and verify that we detect the image.

pkg/config/config.go
```diff
@@ -250,6 +250,32 @@ func (vm *VirtualMachine) AddIgnitionFileFromCmdLine(cmdlineOpts string) error {
...
+					return fmt.Errorf("failed to open file %s: %v", imgPath, err)
+				}
+				defer file.Close()
+				header := make([]byte, 4)
+				_, err = file.Read(header)
+				if err != nil {
+					return fmt.Errorf("failed to read the header of file %s: %v", imgPath, err)
+				}
+				if string(header) == "QFI\xfb" {
+					return fmt.Errorf("vfkit does not support qcow2 block devices")
```
nirs, line 271: qcow2 is a disk image format, not a type of block device.

pkg/config/config.go
```diff
@@ -250,6 +250,32 @@ func (vm *VirtualMachine) AddIgnitionFileFromCmdLine(cmdlineOpts string) error {
 	return nil
 }
 
+func (vm *VirtualMachine) ValidBlkDevices() error {
```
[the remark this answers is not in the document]
nirs: This is a good advice. Even better is to add a Validate() interface to devices so we can keep device related code in the same place.

pkg/config/config_test.go
```diff
@@ -44,3 +47,23 @@ func TestNetworkBlockDevice_NoDevice(t *testing.T) {
...
+
+func TestVirtualMachine_ValidateBlockDevices(t *testing.T) {
+	vm := &VirtualMachine{}
+
+	tmpDir := t.TempDir()
+	imagePath := filepath.Join(tmpDir, "disk.qcow2")
+	size := "1G"
+
+	cmd := exec.Command("qemu-img", "create", "-f", "qcow2", imagePath, size)
+	err := cmd.Run()
```
cfergeau, line 59: Note that this will fail if qemu-img is not installed on the machine. We can improve this in a follow-up PR if this turns out to be a big issue.

pkg/config/config_test.go
```diff
@@ -44,3 +47,23 @@ func TestNetworkBlockDevice_NoDevice(t *testing.T) {
...
+	require.NoError(t, err)
+
+	dev, err := VirtioBlkNew(imagePath)
+	require.NoError(t, err)
+	vm.Devices = append(vm.Devices, dev)
+	defer os.Remove(imagePath)
+
+	err = vm.ValidateDevices()
+	require.Error(t, err)
```
cfergeau, line 68: Can you add a check that the error message is about a qcow2 error?

pkg/config/virtio.go
```diff
@@ -551,6 +562,24 @@ func (dev *VirtioBlk) FromOptions(options []option) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to open file %s: %v", imgPath, err)
+	}
+	defer file.Close()
+	header := make([]byte, 4)
+	_, err = file.Read(header)
+	if err != nil {
+		return fmt.Errorf("failed to read the header of file %s: %v", imgPath, err)
+	}
+	if string(header) == "QFI\xfb" {
```
cfergeau, line 577: A bit worried of the conversion from bytes to string when we have no guarantee the data will be valid utf-8.
I’d favour `if header == []byte("QFI\xfb")` (not sure about the exact syntax)

pkg/config/virtio.go
```diff
@@ -778,6 +819,10 @@ func (nbd *NetworkBlockDevice) FromOptions(options []option) error {
 	return nbd.NetworkBlockStorageConfig.FromOptions(unhandledOpts)
 }
 
+func (nbd *NetworkBlockDevice) Validate() error {
```
cfergeau, line 822: Since `NetworkBlockDevice` embeds a `DiskStorageConfig`, it  might be possible to omit this one.

pkg/config/virtio.go
```diff
@@ -700,6 +737,10 @@ func (dev *RosettaShare) FromOptions(options []option) error {
 	return nil
 }
 
+func (dev *RosettaShare) Validate() error {
```
cfergeau, line 740: I think this could be implemented at the `DirectorySharingConfig` level?

pkg/config/virtio.go
```diff
@@ -655,6 +688,10 @@ func (dev *VirtioFs) FromOptions(options []option) error {
 	return nil
 }
 
+func (dev *VirtioFs) Validate() error {
```
cfergeau, line 691: Same comment as for RosettaShare.

pkg/config/config.go
```diff
@@ -250,6 +250,15 @@ func (vm *VirtualMachine) AddIgnitionFileFromCmdLine(cmdlineOpts string) error {
 	return nil
 }
 
+func (vm *VirtualMachine) ValidateDevices() error {
+	for _, device := range vm.Devices {
+		if err := device.Validate(); err != nil {
```
cfergeau, line 255: There are a number of calls to `Validate` in virtio.go, are they made redundant by this new method? or do we need both?

cfergeau (commented):
Can you split this PR in at least 2 commits, one adding the `Validate` method to the `VirtioDevice` interface (+ValidateDevices), and a second commit adding the qcow2 check? I’d even move `ValidateDevices` to its own commit so that each commit does a single thing.

pkg/config/config_test.go
nirs: We can skip this test if qemu-img is not installed. It is about better integration with qemu, not core vfkit feature. In CI qemu-img can be installed using the package manager.

pkg/config/virtio.go
```diff
@@ -371,7 +378,7 @@ func (dev *VirtioGPU) FromOptions(options []option) error {
 		dev.Height = defaultVirtioGPUResolutionHeight
 	}
 
-	return dev.validate()
+	return dev.Validate()
```
nirs, line 381: This looks like the right way to validate - after creating from options. Maybe we don't need a public Validate() but add the missing validation in VirtioBlk?

nirs (commented):
Looks much nicer like this!

pkg/config/config.go
vyasgun: I think either the Validate() calls at the end of `FromOptions` in some places can be removed or we can add dev.Validate to the end of all `FromOptions`. I have removed the redundant occurrences in this iteration.

pkg/config/virtio.go
vyasgun: We cannot omit this one as `NetworkBlockDevice` has other fields also:

```
type NetworkBlockDevice struct {
	NetworkBlockStorageConfig
	DeviceIdentifier    string
	Timeout             time.Duration
	SynchronizationMode NBDSynchronizationMode
}
```

pkg/config/config.go
cfergeau: One concern I have with this is that this weakens the `FromOptions` and `AddDevicesFromCommandLine` APIs, which are kind of public: I expect other go programs to be using `github.com/crc-org/pkg/config`, even if I’m not sure these specific methods will be widely used.
Before this change, they were returning an error when the config does not validate, after this change, all users of the API need to remember to do the validation themselves or they could get invalid device configuration.

I would keep the validation at the end of each `FromOptions`, or add the validation to either `AddDevicesFromCmdLine` or `deviceFromCmdLine` if we want to avoid the duplication.

vyasgun: Ah yes, that makes sense. I will add validate back to FromOptions

pkg/config/virtio.go
cfergeau: Adding it to the public interface means we need placeholders which return `nil` for 5 to 10 devices, this can be an indication that it does not fit that well in the  interface.
Ideally, the user would never need to call `Validate` explicitly, but it’s going to be called internally by the other public methods, in which can it can be private.

vyasgun:
@cfergeau addressed the comments.

pkg/config/virtio_test.go
```diff
@@ -259,6 +267,22 @@ var virtioDevTests = map[string]virtioDevTest{
 	},
 }
 
+func getTestVirtioBlkDevice() (*VirtioBlk, error) {
+	testImage, err := os.Create(testImagePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create test image %s: %v", testImagePath, err)
+	}
+	if _, err := testImage.Write([]byte{'0', '0', '0', '0'}); err != nil {
```
cfergeau: This could use https://pkg.go.dev/os#WriteFile

pkg/config/virtio_test.go
```diff
@@ -259,6 +267,22 @@ var virtioDevTests = map[string]virtioDevTest{
...
+		testImage.Close()
```
cfergeau: fwiw, I would have used `defer testImage.Close()`, possibly with an error check instead of calling it explicitly, but see the comment about `WriteFile`.

pkg/config/virtio_test.go
```diff
@@ -16,23 +20,27 @@ type virtioDevTest struct {
 	errorMsg         string
 }
 
+var testImagePath = filepath.Join(os.TempDir(), "vfkit_test_disk.img")
```
cfergeau: golang testing package also provides  https://pkg.go.dev/testing#T.TempDir , which has the benefit of being cleaned up automatically. But I can’t think of an easy way of using it given the way our tests are implemented, so your approach is fine.

cfergeau (commented):
Only a few minor comments left

pkg/config/virtio_test.go
nirs: We can create the temp dir at the start of the test using the tests table instead of a global.

pkg/config/virtio_test.go
nirs: It should

pkg/config/virtio_test.go
```diff
@@ -296,6 +320,12 @@ func testErrorVirtioDev(t *testing.T, test *virtioDevTest) {
 }
 
 func TestVirtioDevices(t *testing.T) {
+	defer func() {
+		err := os.Remove(testImagePath)
+		if err != nil {
+			log.Warn("Failed to remove test image")
+		}
+	}()
```
nirs: This can replaced with:

    testImagePath := filepath.Join(t.TempDir(), "test.img")

pkg/config/virtio_test.go
cfergeau: The tricky part is that we are using a table to define the tests (`virtioDevTests`), and some parts of the table need to know the filename: https://github.com/crc-org/vfkit/blob/16d5ddd714db62cf1bf21a2cdc1431021dd9a808/pkg/config/virtio_test.go#L31

But it looks like we can move the table (`virtioDevTests`) to `TestVirtioDevices` where we can define `testImagePath` locally using `t`, and then use this in the table.

pkg/config/virtio_test.go
```diff
@@ -296,6 +65,257 @@ func testErrorVirtioDev(t *testing.T, test *virtioDevTest) {
...
+			expectedCmdLine: []string{"--device", "nbd,uri=nbd://1.1.1.1:10000,timeout=1000,sync=none"},
+		},
+		"NewVirtioBalloon": {
+			newDev:          VirtioBalloonNew,
+			expectedDev:     &VirtioBalloon{},
+			expectedCmdLine: []string{"--device", "virtio-balloon"},
+		},
+	}
+	defer func() {
+		err := os.Remove(testImagePath)
```
cfergeau: One of the benefit of using `t.TempDir()` is that you don’t need to do this cleanup explicitly: 
https://pkg.go.dev/testing#T.TempDir
> TempDir returns a temporary directory for the test to use. The directory is automatically removed when the test and all its subtests complete.

Issue #284, opened 2025-03-15
Sending termination signal does not trigger graceful shutdown


nirs:
When receiving a termination signal (SIGTERM, SIGINT) vfkit runs the exit handlers (removing sockets and iso) and invoke os.Exit(). This stop the guest immediately without doing graceful shutdown, losing guest data that was not flushed to disk.

I think the expected behavior is to do graceful shutdown when terminated normally. A program defaults should be safe and secure. Users that have special needs can use the HTTP API to do a HardStop or kill vfkit.

A side effect of this issue is that we don't run exit handlers then the guest terminates #248, because we have 3 different ways to terminate vfkit (Stop, HardStop, signal).

cfergeau:
See https://github.com/crc-org/vfkit/pull/283#discussion_r1996434522 for some comments on this.

PR #316, opened 2025-06-02
refactor: replace direct os.Exit with graceful shutdown via channel

vyasgun:
Instead of calling os.Exit(1) directly in signal handler, send signal through exitChan to allow proper VM cleanup in main goroutine.

Fixes: https://github.com/crc-org/vfkit/issues/284


vyasgun:
@sourcery-ai review

cmd/vfkit/main.go
```diff
@@ -157,6 +157,13 @@ func runVFKit(vmConfig *config.VirtualMachine, opts *cmdline.Options) error {
 		}
 		srv.Start()
 	}
+	util.RegisterExitHandler(func() {
+		log.Debugf("exiting vfkit")
+		err := vfVM.Stop()
+		if err != nil {
+			log.Error(err)
+		}
+	})
```
cfergeau: I don’t think we want to always run `vfVM.Stop()` in all cases when we exit vfkit.
As Nir summarized, there are (at least) 3 ways to exit vfkit
1) guest stops and vfkit exits
2) the REST API is used to stop/hard stop the VM
3) a signal is sent to the vfkit process

It’s only for 3) that we want to try to run `vfVM.Stop()`, ie run it somehow here https://github.com/crc-org/vfkit/blob/b67809fefca30114d67780864de2501b824a9de3/pkg/util/exithandler.go#L42-L48 and don’t call `ExecuteExitHandler`/`os.Exit()`, but instead let the VM stop, and then the cleanup at exit for 3) can also be done by this code: https://github.com/crc-org/vfkit/blob/b67809fefca30114d67780864de2501b824a9de3/cmd/vfkit/root.go#L34

For 3), maybe we also need to do hardstop after a timeout if the VM does not exit in a reasonable time?

vyasgun: I updated the PR so that there is a channel which captures when exitHandlers are executing and `doExit` is set. Upon receiving the signal, it will `RequestStop` the VM and a hardstop if the state of the vm doesn't change to `Stopped` after 5 seconds.

vyasgun:
@sorcery-ai summary

cmd/vfkit/main.go
```diff
@@ -157,6 +158,25 @@ func runVFKit(vmConfig *config.VirtualMachine, opts *cmdline.Options) error {
 		}
 		srv.Start()
 	}
+	go func() {
+		<-exitChan
```
cfergeau: Might be worth a comment to explain this will get unblocked when SIGKILL/… is received, otherwise this code is not triggered.

pkg/util/exithandler.go
```diff
@@ -28,22 +28,26 @@ func RegisterExitHandler(handler func()) {
...
 	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM, syscall.SIGINT)
 	go func() {
+		defer func() {
+			signal.Stop(sigChan)
+			close(exitChan)
+		}()
 		for sig := range sigChan {
 			log.Printf("captured %v, calling exit handlers and exiting..", sig)
 			ExecuteExitHandlers()
 			if doExit {
```
cfergeau: I think `doExit` is now redundant, you can pass an `exitChan` in the tests, but never try to read from it. Or you can check something is available in `exitChan` after you sent a signal.

pkg/util/exithandler.go
```diff
@@ -28,22 +28,26 @@ func RegisterExitHandler(handler func()) {
...
 	sigChan := make(chan os.Signal, 2)
 	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM, syscall.SIGINT)
 	go func() {
+		defer func() {
+			signal.Stop(sigChan)
+			close(exitChan)
+		}()
 		for sig := range sigChan {
 			log.Printf("captured %v, calling exit handlers and exiting..", sig)
 			ExecuteExitHandlers()
```
cfergeau: Is this still needed? My feeling is that it’s redundant now?

pkg/util/exithandler.go
```diff
@@ -28,22 +28,26 @@ func RegisterExitHandler(handler func()) {
...
-	setupExitSignalHandling(true)
+func SetupExitSignalHandling(exitChan chan bool) {
+	setupExitSignalHandling(exitChan, true)
 }
 
 // setupExitSignalHandling sets up a signal channel to listen for termination or interruption signals.
 // When one of these signals is received, all the registered exit handlers will be invoked.
 // It is possible to prevent the program from exiting by setting the doExit param to false (used for testing)
-func setupExitSignalHandling(doExit bool) {
+func setupExitSignalHandling(exitChan chan bool, doExit bool) {
```
cfergeau: An alternative to this `exitChan` approach would be to pass a `handler func()` similar to what is done with `RegisterExitHandler`. But both approaches are fine, mentioning it here for completeness, changes are not required.

pkg/util/exithandler.go
vyasgun: I don't understand why it might be redundant because closing channels when the function returns will free up resources and not cause any unnecessary leaks.

cfergeau: I was asking about the call to `ExecuteExitHandlers()`. With the changes you made, I expect https://github.com/crc-org/vfkit/blob/5e9c855c06c9afd92f6a56eec0597cf3f683ae0f/cmd/vfkit/root.go#L34 will be triggered even if we exit because of a signal.

vyasgun: ah okay my bad 😅  yes, that's correct but the exit handlers are being called after requesting stop on the VM if we are leaving it upto defer. Added a debug log for reference:
```
~/work/vfkit/out/vfkit --cpus 2 --memory 2048 --bootloader efi,variable-store=out/efistore.nvram,create --device virtio-blk,path=/Users/gvyas/work/vfkit/out/fedora-coreos-41.20250302.3.2-applehv.aarch64.raw --restful-uri http://127.0.0.1:8080
INFO[0000] &{2 2048    {[efi variable-store=out/efistore.nvram create] true}  [virtio-blk,path=/Users/gvyas/work/vfkit/out/fedora-coreos-41.20250302.3.2-applehv.aarch64.raw] http://127.0.0.1:8080  false  {[] false}}
INFO[0000] boot parameters: &{EFIVariableStorePath:out/efistore.nvram CreateVariableStore:true}
INFO[0000]
INFO[0000] virtual machine parameters:
INFO[0000] 	vCPUs: 2
INFO[0000] 	memory: 2048 MiB
INFO[0000]
INFO[0000] Adding virtio-blk device (imagePath: /Users/gvyas/work/vfkit/out/fedora-coreos-41.20250302.3.2-applehv.aarch64.raw)
INFO[0000] virtual machine is running
INFO[0000] waiting for VM to stop
^C2025/06/12 16:14:21 captured interrupt, calling exit handlers and exiting..
WARN[0005] failed to wait for VM stop: timeout waiting for VM state, forcing stop
INFO[0005] VM is stopped
DBG EXECUTING EXIT HANDLERS
```

cfergeau: > yes, that's correct but the exit handlers are being called after requesting stop on the VM if we are leaving it upto defer.

I don’t think it’s a problem when they are called as long as they are called?

pkg/util/exithandler.go
vyasgun: I changed the code so it accepts a shutdown function and executes it. This makes testing also easier as I can provide a test implementation of the shutdown func

PR #317, opened 2025-06-03
Add gvproxy/helper-vmnet scripts

cfergeau:


doc/usage.md
```diff
@@ -311,6 +311,8 @@ This adds a virtio-net device to the VM, and redirects all the network traffic o
 ```
 This is useful in combination with usermode networking stacks such as [gvisor-tap-vsock](https://github.com/containers/gvisor-tap-vsock).
 
+See [this shell script](https://github.com/nirs/vmnet-helper/blob/main/examples/vfkit.sh) for an example of networking using `vmnet-helper`.
```
cfergeau, line 314: forgot to add a link to contrib/scripts/start-gvproxy.sh in the next commit, I’ll do that tomorrow.

contrib/scripts/start-gvproxy.sh
```diff
@@ -0,0 +1,34 @@
...
+#
+# This script creates an overlay for the disk image, the disk image is not modified.
+#
+set -exu
+
+DISK_IMAGE="${1?Usage: $0 diskimage}"
+VM_NAME="$(basename ${DISK_IMAGE})"
+
+TO_REMOVE="${VM_NAME}.sock"
+/usr/local/bin/gvproxy -mtu 1500 -ssh-port 2223 -listen-vfkit unixgram://$(pwd)/${VM_NAME}.sock  -log-file ${VM_NAME}.gvproxy.log &
```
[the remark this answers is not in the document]
cfergeau: This is an example, so I don’t mind if the path needs to be adjusted. However the suggestion makes sense, so we can change it.

contrib/scripts/start-gvproxy.sh
```diff
@@ -0,0 +1,34 @@
...
+# This script can be used to start a raw disk image with vfkit using gvproxy
+# for usermode networking.
+# The mac address must be 5a:94:ef:e4:0c:ee as this is the address expected by gvproxy.
+# If the path to `--listen-vfkit` is too long (more than ~100 characters), then
+# gvproxy/vfkit will fail to start as a unix socket filename must be less than
+# that.
+#
+# This script creates an overlay for the disk image, the disk image is not modified.
+#
+set -exu
```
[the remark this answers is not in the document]
cfergeau: The script does not use pipes, but this is standard enough that we can add it anyway…

contrib/scripts/start-gvproxy.sh
```diff
@@ -0,0 +1,34 @@
...
+
+DISK_IMAGE="${1?Usage: $0 diskimage}"
+VM_NAME="$(basename ${DISK_IMAGE})"
+
+TO_REMOVE="${VM_NAME}.sock"
+/usr/local/bin/gvproxy -mtu 1500 -ssh-port 2223 -listen-vfkit unixgram://$(pwd)/${VM_NAME}.sock  -log-file ${VM_NAME}.gvproxy.log &
+trap 'rm -f ${TO_REMOVE}' EXIT
```
lstocchi: Will this just remove `${VM_NAME}.sock`? Because `TO_REMOVE` gets updated after this point but the rm action will just use the valua of TO_REMOVE here, even if, at the end, the TO_REMOVE contains the img and nvram. Shouldn't we have
`
TO_REMOVE="${VM_NAME}.sock ${VM_NAME}.overlay.img ${VM_NAME}.efistore.nvram"
trap 'rm -f ${TO_REMOVE}' EXIT
`
??

cfergeau: I was not sure about the behaviour when working on the PR, but I tested it, and `${TO_REMOVE}` is evaluated at the time the `trap` triggers (ie right before exiting, after it’s been updated), not at the time the `trap` is defined.

However, I considered writing it your way for more clarity, I’ll change it to avoid the next person reading this code being unsure.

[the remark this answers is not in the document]
cfergeau: Done

contrib/scripts/start-gvproxy.sh
```diff
@@ -0,0 +1,39 @@
...
+: "${VFKIT:=./out/vfkit}"
+
+DISK_IMAGE="${1?Usage: $0 diskimage}"
+VM_NAME="$(basename ${DISK_IMAGE})"
+
+TO_REMOVE="${VM_NAME}.sock ${VM_NAME}.gvproxy.pid ${VM_NAME}.overlay.img ${VM_NAME}.efistore.nvram"
+${GVPROXY} -mtu 1500 -ssh-port 2223 -listen-vfkit unixgram://$(pwd)/${VM_NAME}.sock  -log-file ${VM_NAME}.gvproxy.log --pid-file ${VM_NAME}.gvproxy.pid &
+trap 'if [[ -f "${VM_NAME}.gvproxy.pid" ]]; then kill $(cat ${VM_NAME}.gvproxy.pid); fi; rm -f ${TO_REMOVE}' EXIT
+
+TO_REMOVE="${TO_REMOVE} ${VM_NAME}.overlay.img"
```
lstocchi: You forgot to remove this and line 33

lstocchi (commented):
/lgtm

contrib/scripts/start-gvproxy.sh
```diff
@@ -0,0 +1,38 @@
...
+#
+set -exuo pipefail
+
+: "${GVPROXY:=gvproxy}"
+: "${VFKIT:=./out/vfkit}"
+
+DISK_IMAGE="${1?Usage: $0 diskimage}"
+VM_NAME="$(basename ${DISK_IMAGE})"
+
+${GVPROXY} -mtu 1500 -ssh-port 2223 -listen-vfkit unixgram://$(pwd)/${VM_NAME}.sock  -log-file ${VM_NAME}.gvproxy.log --pid-file ${VM_NAME}.gvproxy.pid &
```
nirs, line 26: This command mixes single dash `-log-file` and double dash `--pid-file` even if it works, we should use the same style for consistency.

PR #327, opened 2025-06-24
cmd: add support for nested virtualization

vyasgun:
This commit adds a --nested-virtualization (-n) flag to enable nested virtualization in vfkit. When enabled on macOS, it sets up a vz.GenericPlatformConfiguration with SetNestedVirtualizationEnabled(true).


pkg/cmdline/cmdline.go
```diff
@@ -56,4 +58,5 @@ func AddFlags(cmd *cobra.Command, opts *Options) {
 
 	cmd.Flags().StringVar(&opts.IgnitionPath, "ignition", "", "path to the ignition file")
 	cmd.Flags().VarP(&opts.CloudInitFiles, "cloud-init", "", "path to user-data and meta-data cloud-init configuration files")
+	cmd.Flags().BoolVarP(&opts.NestedVirtualization, "nested-virtualization", "n", false, "use nested virtualization")
```
gbraad: Nitpicking, but this is not 'use' nested virtualization, but 'enable'.

This is also named as such; `platformConfig.SetNestedVirtualization*Enabled*(true)`

pkg/cmdline/cmdline.go
```diff
@@ -56,4 +58,5 @@ func AddFlags(cmd *cobra.Command, opts *Options) {
 
 	cmd.Flags().StringVar(&opts.IgnitionPath, "ignition", "", "path to the ignition file")
 	cmd.Flags().VarP(&opts.CloudInitFiles, "cloud-init", "", "path to user-data and meta-data cloud-init configuration files")
+	cmd.Flags().BoolVarP(&opts.NestedVirtualization, "nested-virtualization", "n", false, "enable nested virtualization")
```
cfergeau: The flag should be `--nested` for compatibility with krunkit.

pkg/vf/vm.go
```diff
@@ -39,13 +39,39 @@ func NewVirtualMachine(vmConfig config.VirtualMachine) (*VirtualMachine, error)
 		vfConfig.SetPlatformVirtualMachineConfiguration(platformConfig)
 	} else {
 		PlatformType = "linux"
+		if vmConfig.NestedVirtualization {
+			err := setNestedVirtualization(vfConfig)
```
cfergeau: Can we take a similar approach to the `macosBootloader` branch and introduce a `NewGenericPlatformConfiguration` function? At the moment, nested virtualization won’t be enabled with macOS guests.

cfergeau:
> I’ll test this on a machine without nested support

This fails with this message which should be ok
```
INFO[0000] Adding virtio-rng device
Error: Error Domain=VZErrorDomain Code=2 Description="Invalid virtual
machine configuration. Nested virtualization is not available on this
hardware." UserInfo={
     NSLocalizedFailure = "Invalid virtual machine configuration.";
     NSLocalizedFailureReason = "Nested virtualization is not available
on this hardware.";
```

pkg/vf/vm.go
vyasgun: do you mean replace `NewMacPlatformConfiguration` with `NewGenericPlatformConfiguration`?

gbraad:
Lot of duplication in that error (`INFO`) message. Does it have a specific error code, or a static message that can be interpreted? If vfkit is started in the background, this would otherwise not be recorded by the calling application.

cfergeau:
> Lot of duplication in that error (`INFO`) message. Does it have a specific error code, or a static message that can be interpreted? If vfkit is started in the background, this would otherwise not be recorded by the calling application.

`Nested virtualization is not available on this hardware` can be used as a static string for a match. Otherwise we can come up with our own error message by using https://github.com/Code-Hex/vz/blob/5bbaec6daefb962461fdf7fb33b27aff8e939cff/platform.go#L44-L51

pkg/vf/vm.go
cfergeau: The `macOSBootloader` branch has to keep using `NewMacPlatformConfiguration`, but the linux codepath could make use of `NewGenericPlatformConfiguration` to have some kind of symmetry between the 2.

gbraad:
> by using ...

This is merely based on the macOS version, so not a hardware limitation? Does this mean 15 on an M1 would support this?

cfergeau:
> > by using ...
> 
> This is merely based on the macOS version, so not a hardware limitation? Does this mean 15 on an M1 would support this?

It’s not merely based on the macOS version. If you have macOS < 15, then nested virt is not supported. Otherwise, on a M1,  this would return false:
```
	return (bool)(C.isNestedVirtualizationSupported())
```

vyasgun:
I added a check for availability of nested virtualization like this:
```
if opts.Nested && !vz.IsNestedVirtualizationSupported() {
		return nil, fmt.Errorf("nested virtualization is not supported")
	}
```

pkg/vf/vm.go
vyasgun: ah, okay got it. I'll add the function

gbraad:
When requested, but not available, would you error out or only inform the user (but continue)?

If have an opinion, but WDYT?

vyasgun:
@gbraad would it be correct to continue with the `--nested` flag if it's not going to be supported inside the VM?

gbraad:
@cfergeau WDYS?

cfergeau:
> @cfergeau WDYS?

If someone requested nested virtualization when it’s not available, this should be an error.
If that causes problems in some use cases, we can add a way to check if it’s available from the commandline, or a flag to ignore the error, …  But by default, it’s an error to ask for nested virt and not to get it.

gbraad:
Agree, fail early... error out hard. You can not continue when something you requested is not available.

pkg/vf/vm_amd64.go
```diff
@@ -18,3 +18,7 @@ func toVzMacOSBootloader(_ *config.MacOSBootloader) (vz.BootLoader, error) {
 func newMacGraphicsDeviceConfiguration(_ *VirtioGPU) (vz.GraphicsDeviceConfiguration, error) {
 	return nil, fmt.Errorf("running macOS guests is only supported on ARM devices")
 }
+
+func NewGenericPlatformConfiguration(vmConfig config.VirtualMachine) (vz.PlatformConfiguration, error) {
+	return nil, fmt.Errorf("running generic platform configuration is only supported on ARM devices")
```
cfergeau: Won’t this prevent us from starting VMs on intel? The "platform" API seems available on intel, was there a problem with having `NewGenericPlatformConfiguration` in `vm.go`?

pkg/vf/vm_arm64.go
```diff
@@ -47,6 +47,27 @@ func NewMacPlatformConfiguration(machineIdentifierVar, hardwareModelVar, auxilia
...
+	platformConfig, err := vz.NewGenericPlatformConfiguration(
+		vz.WithGenericMachineIdentifier(identifier),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error creating generic platform configuration: %w", err)
+	}
+
+	if vmConfig.Nested {
+		err = platformConfig.SetNestedVirtualizationEnabled(true)
+		if err != nil {
```
cfergeau: Can this part be moved out of `NewGenericPlatformConfiguration` so that we can enable nested virt both for mac and linux VMs?

pkg/vf/vm_amd64.go
vyasgun: I tried on an Intel machine and virtualization was not available on it so I thought it's only there on arm64 machines. I will put the function in `vm.go` as that's better

pkg/vf/vm_arm64.go
vyasgun: I'm not sure how to do this since the method `SetNestedVirtualizationEnabled` is only available in `GenericPlatformConfiguration`

cfergeau: Oh indeed, I missed that! We probably should return an error in the macos code path if nested virt was requested.
Apparently no nested virt with macos guests? https://github.com/utmapp/UTM/issues/6700#issuecomment-2533720620 this is unexpected!

vyasgun: Returning an error for macOS when nested virt is requested sounds like the right thing to do. Made the change.

Also, I didn’t expect nested virt to be unsupported for macOS guests either. That UTM issue was helpful, thanks for linking it!

gbraad: maybe implied, always possible... or not? would be good to verify/look into further. at least something to describe as I am sure this will come up as a question

Edit: read the issue, and.seems clear; no virt inside nested macos. documentation issue in that case and I would say; either error out when requested on mac or, in this case an INFO as there is nothing you can do about this.

vyasgun: I tried with tart and it gives an error that nested virtualization is unsupported on macOS guests.

gbraad: ~This is important to also note on https://github.com/crc-org/crc-internal/issues/123, as that would conclude it; the idea was to use macOS to run nested virtualization to test vfkit in a 'clean host'.~ Done

PR #348, opened 2025-08-21
ci: add macOS-15 runner and enable tests on all macOS versions

vyasgun:


cfergeau:
I don’t know if macos-15 runners can run VMs, I hope so :-/

cfergeau:
The comments from coderabbit about the workers’ naming (`macOS-15` vs `macos-15`) are in my opinion worth addressing

nirs:
> I don’t know if macos-15 runners can run VMs, I hope so :-/

They can not run vms since they are M1 machines. Nested virtualization requires M3 machines. Github provides M1 machines for standard runners and M2 machines for large runners.

- https://docs.github.com/en/enterprise-cloud@latest/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories
- https://docs.github.com/en/enterprise-cloud@latest/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-internal-and-private-repositories
- https://docs.github.com/en/enterprise-cloud@latest/actions/reference/runners/larger-runners

.github/workflows/compile.yml
```diff
@@ -25,6 +25,7 @@ jobs:
         os:
         - macOS-13
```
nirs: You can run vms only on macos-13 (intel)

.github/workflows/compile.yml
```diff
@@ -25,6 +25,7 @@ jobs:
...
         - macOS-14
+        - macOS-15
```
nirs: macos-14 and macos-15 are fine for running unit test but you cannot run vms in the tests since the runner is a vm, and nested virtualization is not supported on the GitHub runners.

Makefile
```diff
@@ -11,8 +11,12 @@ include tools/tools.mk
 
 build: out/vfkit
 
-test: build
+test: test-unit test-integration
+
+test-unit:
 	@go test -v ./pkg/...
+
```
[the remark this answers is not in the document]
vyasgun: but this would change the initial configuration while i only want to split into unit and integration targets

.github/workflows/compile.yml
```diff
@@ -52,14 +53,18 @@ jobs:
...
         run: qemu-img --version
 
-      - name: Test
-        if: matrix.os != 'macOS-14'
-        run: make test
+      - name: Unit tests
+        run: make test-unit
+
+      - name: Integration tests
+        if: matrix.os == 'macOS-13'
```
nirs: macos-13 (lowercase)

.github/workflows/compile.yml
```diff
@@ -52,14 +53,18 @@ jobs:
...
       - name: vet
         run: go vet ./...
       - name: Upload vfkit artifact
-        if: matrix.os == 'macOS-14'
+        if: matrix.os == 'macos-15'
         uses: actions/upload-artifact@v4
         with:
           name: Unsigned vfkit Universal Binary
           path: "./out/vfkit"
+          if-no-files-found: error
```
nirs, line 75: Is this related to adding macos-15 runner? seems like another fix that can be in a separate commit with a commit message explaining the issue it fixes.

.github/workflows/compile.yml
```diff
@@ -25,6 +25,7 @@ jobs:
         os:
```
nirs: not related to this change, but this is really a runner, not an os.

It will look better in:

    runs-on: ${{ matrix.runner }}

nirs (approved):

.github/workflows/compile.yml
vyasgun: this was a suggestion from coderabbit

.github/workflows/compile.yml
vyasgun: Replaced os with runner

nirs (commented):
@vyasgun please address coderabbit comments about the runner names.

Also fix `macOS-15` in the commit message, we want to always use the right name of the runner. Maybe your editor is replacing `macos` with `macOS` automatically? I have the same issue with Safari.

.github/workflows/compile.yml
```diff
@@ -45,21 +46,29 @@ jobs:
...
+        if: matrix.runner == 'macos-15'
         uses: actions/upload-artifact@v4
         with:
           name: Unsigned vfkit Universal Binary
           path: "./out/vfkit"
+          if-no-files-found: error
+      - name: Verify universal binary (optional)
+        if: matrix.runner == 'macos-15'
+        run: |
+          command -v lipo >/dev/null && lipo -info ./out/vfkit || file ./out/vfkit
```
nirs: This change is not related to adding macos-15 runner. Please split it to another commit with a commit message explaining the change. Adding unrelated changes make this harder to review and makes reverting problematic changes harder. We want to have small commits, each doing one logic change.

Regarding this change, this shows information about the executable and it does not verify anything. The name of the step is misleading.

It is also not clear what is (optional) about this step. If this may fail in CI, we need to handle failures in the script and not let fail. But in general we should keep the CI simple as possible, so adding optional steps is not a good way.

If we build a universal binary with lipo the command cannot be missing, and there is no reason why it would fail to show info about the binary and we need to fallback to the file command. Make this command required and fail the build if it fails.

We should have a nightly build verifying the the workflows work. If at some point lips will start to fail we want to know about it quickly. Making this required step will make it easy to detect.

nirs (changes requested):

.github/workflows/compile.yml
vyasgun: yes, you are right. I just wanted to try the coderabbit suggestion on the github workflow once and see how it goes. It doesn't seem necessary so I will remove this step altogether.

vyasgun:
@nirs thank you for the review! I addressed all the comments. Hope it looks okay.

.github/workflows/compile.yml
```diff
@@ -48,18 +49,20 @@ jobs:
...
       - name: vet
         run: go vet ./...
       - name: Upload vfkit artifact
-        if: matrix.os == 'macOS-14'
+        if: matrix.runner == 'macos-15'
         uses: actions/upload-artifact@v4
         with:
           name: Unsigned vfkit Universal Binary
           path: "./out/vfkit"
+          if-no-files-found: error
```
nirs: Do we miss a newline at the end of the last line?

nirs (approved):

Makefile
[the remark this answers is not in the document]
cfergeau: It would be good to run this missing test nonetheless, this can be added in a separate commit.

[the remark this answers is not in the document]
cfergeau: And adding to the commit log an explanation *why* we are doing this split would be really useful, as we won’t necessarily remember this in a few months

.github/workflows/compile.yml
```diff
@@ -48,18 +49,20 @@ jobs:
...
       - name: Verify qemu-img is installed
         run: qemu-img --version
+      - name: Unit tests
+        run: make test-unit
 
-      - name: Test
-        if: matrix.os != 'macOS-14'
-        run: make test
+      - name: Integration tests
+        if: matrix.runner == 'macos-13'
```
cfergeau, line 65: The main goal of switching to newer runners was to look into running this test on newer macos versions.
I understand it’s not possible, but this is not explained anywhere in the PR description or the commit logs…
The reason why it’s problematic is https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down
The macos 13 runners will be gone by the end of the year, I don’t know if there will be other runners with virt support to replace them :(

vyasgun: I was checking out the large macos runners: https://docs.github.com/en/actions/reference/runners/larger-runners#runner-images
They run on VMs according to this and since nested virtualisation is not supported, it wouldn't be able to run VMs. I think we might have to go with a self hosted runner after macos-13 support goes away

nirs: We need a comment on the runners - which arch it is using, and which kind of tests we run on it.

vyasgun: Added the comments

nirs (approved):

cfergeau:
Nir just brought https://github.com/actions/runner-images/issues/13045 to my attention. There will be `macos-15-intel` images which hopefully will have virt support.

nirs:
The runners is available now:
https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

cfergeau:
> The runners is available now: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

Indeed https://github.com/cfergeau/vfkit/actions/runs/18099008305/job/51496812139 \o/

nirs:
We can also add macos-26 to the matrix.

Issue #352, opened 2025-08-29
Consider adding --device virtio-net,type=unixgram for krunkit compatibility


nirs:
krunkit added new way to work with unix datagram socket:
https://github.com/containers/krunkit/pull/63

```
--device virtio-net,type=unixgram,path=/path/to/socket.sock,mac=...,offloading=on,vfkitMagic=on
```

This device creates the same configuration as:

```
--device virtio-net,unixSocketPath=/path/to/socket.sock,mac=...
```

But unlike unixSocketPath it
- supports multiple devices
- does not enable offloading by default (7 times faster host-to-vm performance with vment-helper)
- does not send the vfkit magic (VFKT) by default (not needed with vmnet-helper)

krunkit supports also other types like:

```
--device virtio-net,type=unixstream,path=/path/to/socket.sock,mac=...
```

This is required to use passt or socket_vment.

For vfkit type=unixstream is not relevant since the virtualization framework supports only unix datagram socket, but type=unixgrram is matches.

The offloading option is not relevant to vfkit since the virtualization framework does not have a way to enable offloading and using makes performance much worse.

The vfkitMagic flag makes sense since it is needed only for gvproxy and can be removed in future version of gvproxy.

Adding similar configuration will make it easier to use vfkit and krunkit.

## Suggested change

```
--device virtio-net,type=unixgram,path=/path/to/socket.sock,mac=...,vfkitMagic=on
```

- 'type`: the only option is `unixgram`
- `path`: path to unix datagram socket
- `mac`: mac address
- `vfkitMagic`: if specified, send vfkit magic packet (needed for gvproxy)

Internally no change is needed except making the vfkit magic packet optional.

cfergeau:
`VFKT` will be optional in gvisor-tap-vsock once https://github.com/containers/gvisor-tap-vsock/pull/530 merges.
It makes sense to add a `vfkitMagic` boolean to the command line to help with the transition.
Does krunkit also support `unixSocketPath=/path/to/socket.sock`? or only `type=unixgram,path=/path/to/socket.sock`?

nirs:
> Does krunkit also support `unixSocketPath=/path/to/socket.sock`? or only `type=unixgram,path=/path/to/socket.sock`?

`unixSocketPath=/path/to/socket.sock` is supported for backward compatibility with existing users, but the recommended configuration is type=unixgram or type=unixtream since these support both path= and fd= and allow enabling offloading and vfkitMagic which are disabled by default.

[...]

PR #372, opened 2025-09-26
Add krunkit compatibility for virtio-net devices

vyasgun:
Support krunkit-compatible syntax for issue https://github.com/crc-org/vfkit/issues/352:
- type=unixgram: compatibility marker for unix sockets
- path=: alias for unixSocketPath
- vfkitMagic=on/off: controls VFKT magic packet
- offloading=on/off: accepted for compatibility (no-op)

Also modified tests using device virtio-net:
- Update VirtioNet test to expect krunkit format output
- Update JSON test to include VfkitMagic field
- Ensure all tests pass with new ToCmdLine format


pkg/config/virtio.go
```diff
@@ -494,11 +505,35 @@ func (dev *VirtioNet) FromOptions(options []option) error {
...
+			dev.UnixSocketPath = option.value
+			hasPath = true
+			dev.VfkitMagic = true
+		case "vfkitMagic":
+			if option.value != "on" && option.value != "off" {
+				return fmt.Errorf("invalid value for vfkitMagic: %s (expected on/off)", option.value)
+			}
+			dev.VfkitMagic = option.value == "on"
+		case "offloading":
+			if option.value != "on" && option.value != "off" {
```
lstocchi: Based on https://github.com/crc-org/vfkit/issues/352#issuecomment-3279855986 only the "on" value should be accepted

pkg/config/virtio.go
```diff
@@ -494,11 +505,35 @@ func (dev *VirtioNet) FromOptions(options []option) error {
...
+				return fmt.Errorf("invalid value for offloading: %s (expected on/off)", option.value)
+			}
 		default:
 			return fmt.Errorf("unknown option for virtio-net devices: %s", option.key)
 		}
 	}
 
+	// Validate that type and path are used together
+	if hasType && !hasPath {
```
lstocchi: maybe this should go in the validate() func

pkg/config/virtio.go
```diff
@@ -494,11 +505,35 @@ func (dev *VirtioNet) FromOptions(options []option) error {
...
+			dev.VfkitMagic = true
+		case "type":
+			if option.value != "unixgram" && option.value != "" {
+				return fmt.Errorf("unsupported virtio-net type: %s (only 'unixgram' is supported)", option.value)
+			}
+			hasType = true
+		case "path":
+			dev.UnixSocketPath = option.value
+			hasPath = true
+			dev.VfkitMagic = true
```
cfergeau: I’d set `dev.VfkitMagic = true` once before the loop over `options` and not when handling `path` and `unixSocketPath`, otherwise `vfkitMagic=off,unixSocketPath=/path.sock` will silently enable `vfkitMagic`.

pkg/config/virtio_test.go
```diff
@@ -205,8 +205,9 @@ func TestVirtioDevices(t *testing.T) {
 			expectedDev: &VirtioNet{
 				Nat:            false,
 				UnixSocketPath: "/tmp/unix.sock",
+				VfkitMagic:     true,
 			},
-			expectedCmdLine: []string{"--device", "virtio-net,unixSocketPath=/tmp/unix.sock"},
+			expectedCmdLine: []string{"--device", "virtio-net,type=unixgram,path=/tmp/unix.sock,vfkitMagic=on"},
```
cfergeau, line 213: Maybe there could be a few more tests to test `type=foo` or `vfkitMagic=off`?

pkg/config/virtio.go
vyasgun: `type` and `path` are not part of the device struct, and adding them would be redundant since the type is always 'unixgram' and the options `path` and `unixSocketPath` both refer to the same field. This approach only validates the syntax during command line parsing.

pkg/config/virtio.go
vyasgun: yes, but to maintain compatibility, can accept both values but throw an error when `on` is provided.

pkg/config/virtio_test.go
vyasgun: Added

pkg/config/virtio.go
vyasgun: Fixed this

pkg/config/virtio.go
lstocchi: not sure how good it is. At least I would update the error message to only show `off` as valid value. If we mention that also "on" (`(expected on/off)`) is supported, me (as a user) expects that I have the chance to use it.

lstocchi: btw i understand your point for maintaining compatibility (offloading accepts only `on` and `off` as values). I'm just saying that for me it does not make sense to show `on` as a valid value to the user when it cannot be used ...  but it is not a blocker and can be left as it is if everyone is ok with the current state

pkg/config/virtio.go
```diff
@@ -494,11 +512,43 @@ func (dev *VirtioNet) FromOptions(options []option) error {
 			dev.Socket = os.NewFile(uintptr(fd), "vfkit virtio-net socket")
 		case "unixSocketPath":
 			dev.UnixSocketPath = option.value
+		case "type":
+			if option.value != "unixgram" && option.value != "" {
```
cfergeau: Is it possible to get `option.value==""` when `type=` is specified on the command line?

pkg/config/virtio.go
```diff
@@ -473,6 +481,16 @@ func (dev *VirtioNet) ToCmdLine() ([]string, error) {
...
 func (dev *VirtioNet) FromOptions(options []option) error {
+	var hasType bool
+	var typeOnlyOptions []string // Options that require type to be specified
+
+	for _, option := range options {
+		if option.key == "path" || option.key == "unixSocketPath" {
+			dev.VfkitMagic = true
+			break
+		}
+	}
```
cfergeau, line 492: Could this be done unconditionnally without looping? Just `dev.VfkitMagic=true`?

Fwiw, the explicit loop can be avoided by using https://pkg.go.dev/slices#ContainsFunc
You can keep the code as it is as it’s likely a matter of preference to use one or the other.

pkg/config/virtio.go
```diff
@@ -494,11 +512,43 @@ func (dev *VirtioNet) FromOptions(options []option) error {
...
+	// Validate type+path dependency and type-only options
+	if hasType && dev.UnixSocketPath == "" {
+		return fmt.Errorf("'type' option requires 'path' to be specified")
+	}
+
+	for _, optName := range typeOnlyOptions {
+		if !hasType {
+			return fmt.Errorf("'%s' option requires 'type' to be specified", optName)
+		}
+	}
```
cfergeau, line 545: Imo it’s fine to use `unixSocketPath=…,vfkitMagic=off` for example. The more modern approach is preferred, but I don’t see a reason for forcing its use to use the new feature.

pkg/config/virtio_test.go
```diff
@@ -308,6 +309,97 @@ func TestVirtioDevices(t *testing.T) {
 			expectedDev:     &VirtioBalloon{},
 			expectedCmdLine: []string{"--device", "virtio-balloon"},
 		},
+		"VirtioNetWithVfkitMagicOff": {
+			newDev: func() (VirtioDevice, error) {
+				return deviceFromCmdLine("virtio-net,type=unixgram,path=/tmp/test.sock,vfkitMagic=off")
```
cfergeau: Do we have tests checking that parsing `unixSocketPath=…` does the right thing?

pkg/config/virtio.go
vyasgun: Changed it to use ContainsFunc

pkg/config/virtio.go
vyasgun: okay, I will make `on` an invalid value. `vfkit` would error out anyway if offloading=on is provided

pkg/config/virtio.go
```diff
@@ -494,11 +512,43 @@ func (dev *VirtioNet) FromOptions(options []option) error {
...
+			dev.VfkitMagic = option.value == "on" // This overrides the default
+			typeOnlyOptions = append(typeOnlyOptions, option.key)
+		case "offloading":
+			if option.value != "on" && option.value != "off" {
+				return fmt.Errorf("invalid value for offloading: %s (expected on/off)", option.value)
+			}
+			if option.value == "on" {
+				return fmt.Errorf("offloading is not supported in vfkit")
+			}
+			typeOnlyOptions = append(typeOnlyOptions, option.key)
```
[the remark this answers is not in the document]
vyasgun: offloading is a no-op. vfkitMagic is used to toggle VfkitMagic field.

pkg/config/virtio.go
vyasgun: yes, it's possible but I think it's better to accept unixgram explicitly so I changed this.

pkg/config/virtio_test.go
vyasgun: Added

pkg/config/virtio.go
vyasgun: Added the changes and tests also

pkg/config/virtio.go
```diff
@@ -494,11 +512,39 @@ func (dev *VirtioNet) FromOptions(options []option) error {
...
 		}
 	}
 
+	// Validate type+path dependency and type-only options
+	if hasType && dev.UnixSocketPath == "" {
+		return fmt.Errorf("'type' option requires 'path' to be specified")
+	}
+
+	for _, optName := range typeOnlyOptions {
+		if !hasType {
```
cfergeau: I’d rewrite this one as 
```
if !hasType && len(typeOnlyOptions) != 0 {
    return fmt.Errorf("'%s' option requires 'type' to be specified", typeOnlyOptions[0])
}
```
as we’ll never go further than the first option in the `typeOnlyOptions` slice, no need for a loop.

PR #373, opened 2025-09-28
feat: Support raw block devices

mcfx:
This PR adds support for raw block devices.

Since there is `NewDiskBlockDeviceStorageDeviceAttachment` API in vz, I was able to implement it with very little code.

However, I’m not sure if this is the best approach. Perhaps adding a new config type would be better? Or maybe we should warn users that they are performing a potentially dangerous operation? Suggestions are welcome.

## Summary by CodeRabbit

* **New Features**
  * Per-disk backend selection via a new Type (image or dev) exposed in CLI, JSON, and runtime.

* **Bug Fixes**
  * Explicit validation and error reporting for disk image paths and host block devices; improved device-add logging and preserved image caching/sync behavior.

* **Documentation**
  * Usage docs updated with type parameter, examples for attaching host block devices, and cloud-init notes.

* **Tests**
  * Added tests for backend types and updated JSON stability expectations.

### DIFF
--- a/pkg/config/json_test.go
+++ b/pkg/config/json_test.go
@@ -285,29 +285,32 @@ var jsonStabilityTests = map[string]jsonStabilityTest{
 		newObjectFunc: func(t *testing.T) any {
 			blk, err := VirtioBlkNew("")
 			require.NoError(t, err)
+			blk.Type = DiskBackendImage
 			return blk
 		},
 
-		skipFields:   []string{"DevName", "URI"},
-		expectedJSON: `{"kind":"virtioblk","devName":"virtio-blk","imagePath":"ImagePath","readOnly":true,"deviceIdentifier":"DeviceIdentifier"}`,
+		skipFields:   []string{"DevName", "URI", "Type"},
+		expectedJSON: `{"kind":"virtioblk","devName":"virtio-blk","imagePath":"ImagePath","readOnly":true,"type":"image","deviceIdentifier":"DeviceIdentifier"}`,
 	},
 	"USBMassStorage": {
 		newObjectFunc: func(t *testing.T) any {
 			usb, err := USBMassStorageNew("")
 			require.NoError(t, err)
+			usb.Type = DiskBackendImage
 			return usb
 		},
-		skipFields:   []string{"DevName", "URI"},
-		expectedJSON: `{"kind":"usbmassstorage","devName":"usb-mass-storage","imagePath":"ImagePath","readOnly":true}`,
+		skipFields:   []string{"DevName", "URI", "Type"},
+		expectedJSON: `{"kind":"usbmassstorage","devName":"usb-mass-storage","imagePath":"ImagePath","readOnly":true,"type":"image"}`,
 	},
 	"NVMExpressController": {
 		newObjectFunc: func(t *testing.T) any {
 			nvme, err := NVMExpressControllerNew("")
 			require.NoError(t, err)
+			nvme.Type = DiskBackendImage
 			return nvme
 		},
-		skipFields:   []string{"DevName", "URI"},
-		expectedJSON: `{"kind":"nvme","devName":"nvme","imagePath":"ImagePath","readOnly":true}`,
+		skipFields:   []string{"DevName", "URI", "Type"},
+		expectedJSON: `{"kind":"nvme","devName":"nvme","imagePath":"ImagePath","readOnly":true,"type":"image"}`,
 	},
 	"LinuxBootloader": {
 		obj:          &LinuxBootloader{},
--- a/pkg/config/virtio.go
+++ b/pkg/config/virtio.go
@@ -902,9 +902,32 @@ type StorageConfig struct {
 	ReadOnly bool   `json:"readOnly,omitempty"`
 }
 
+type DiskBackendType string
+
+const (
+	/// Normal disk images, like .img
+	DiskBackendImage DiskBackendType = "image"
+
+	/// Real block devices, like /dev/disk1s1
+	DiskBackendBlockDevice DiskBackendType = "dev"
+
+	/// If the value is empty, it defaults to image
+	DiskBackendDefault DiskBackendType = ""
+)
+
+func (typ DiskBackendType) IsValid() bool {
+	switch typ {
+	case DiskBackendImage, DiskBackendBlockDevice, DiskBackendDefault:
+		return true
+	default:
+		return false
+	}
+}
+
 type DiskStorageConfig struct {
 	StorageConfig
-	ImagePath string `json:"imagePath,omitempty"`
+	ImagePath string          `json:"imagePath,omitempty"`
+	Type      DiskBackendType `json:"type,omitempty"`
 }
 
 type NetworkBlockStorageConfig struct {
@@ -919,6 +942,10 @@ func (config *DiskStorageConfig) ToCmdLine() ([]string, error) {
 
 	value := fmt.Sprintf("%s,path=%s", config.DevName, config.ImagePath)
 
+	if config.Type != DiskBackendDefault {
+		value += fmt.Sprintf(",type=%s", string(config.Type))
+	}
+
 	if config.ReadOnly {
 		value += ",readonly"
 	}
@@ -930,6 +957,12 @@ func (config *DiskStorageConfig) FromOptions(options []option) error {
 		switch option.key {
 		case "path":
 			config.ImagePath = option.value
+		case "type":
+			typ := DiskBackendType(option.value)
+			if !typ.IsValid() {
+				return fmt.Errorf("unexpected value for disk 'type' option: %s", option.value)
+			}
+			config.Type = typ
 		case "readonly":
 			if option.value != "" {
 				return fmt.Errorf("unexpected value for virtio-blk 'readonly' option: %s", option.value)
--- a/pkg/config/virtio_test.go
+++ b/pkg/config/virtio_test.go
@@ -105,6 +105,49 @@ func TestVirtioDevices(t *testing.T) {
 			expectedCmdLine:  []string{"--device", fmt.Sprintf("virtio-blk,path=%s,deviceId=test", testImagePath)},
 			alternateCmdLine: []string{"--device", fmt.Sprintf("virtio-blk,deviceId=test,path=%s", testImagePath)},
 		},
+		"NewVirtioBlkWithType": {
+			newDev: func() (VirtioDevice, error) {
+				dev, err := getTestVirtioBlkDevice(testImagePath)
+				if err != nil {
+					return nil, err
+				}
+				dev.Type = DiskBackendBlockDevice
+				return dev, nil
+			},
+			expectedDev: &VirtioBlk{
+				DiskStorageConfig: DiskStorageConfig{
+					StorageConfig: StorageConfig{
+						DevName: "virtio-blk",
+					},
+					ImagePath: testImagePath,
+					Type:      DiskBackendBlockDevice,
+				},
+				DeviceIdentifier: "",
+			},
+			expectedCmdLine:  []string{"--device", fmt.Sprintf("virtio-blk,path=%s,type=dev", testImagePath)},
+			alternateCmdLine: []string{"--device", fmt.Sprintf("virtio-blk,type=dev,path=%s", testImagePath)},
+		},
+		"NewVirtioBlkWithDefaultType": {
+			newDev: func() (VirtioDevice, error) {
+				dev, err := getTestVirtioBlkDevice(testImagePath)
+				if err != nil {
+					return nil, err
+				}
+				dev.Type = DiskBackendDefault
+				return dev, nil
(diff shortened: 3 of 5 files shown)


pkg/vf/virtio.go
```diff
@@ -490,6 +490,31 @@ func (config *DiskStorageConfig) toVz() (vz.StorageDeviceAttachment, error) {
...
+	}
+	stat, ok := info.Sys().(*syscall.Stat_t)
+	if !ok {
+		return nil, fmt.Errorf("unable to get raw syscall.Stat_t data")
+	}
+	mode := stat.Mode
+	if mode & syscall.S_IFMT == syscall.S_IFBLK {
+		var mode int
+		if config.ReadOnly {
+			mode = os.O_RDONLY
```
cfergeau: Can you call this variable `flags`, or `open_flags` or such? I initially thought there was a mixup between the 2nd and the 3rd arg of [`OpenFile`](https://pkg.go.dev/os#OpenFile) which is of type `FileMode`

mcfx:
Thank you for your comments.

I agree with that a separate `type=dev` is better, and I will implement in the next few days. I will also add some documents and tests.

I also agree that we should use create a separate root process, and pass the fd. But I'm not familiar with the overall design of vfkit, and I don't know the best way to organize the source code of this separate binary and its user experience. So I dedide to keep using a path directly in this PR.

pkg/vf/virtio.go
mcfx: I will rename it in next commit

cfergeau:
> I also agree that we should use create a separate root process, and pass the fd. But I'm not familiar with the overall design of vfkit, and I don't know the best way to organize the source code of this separate binary and its user experience. So I dedide to keep using a path directly in this PR.

yeah, a path is ok for now, this can be improved later when someone really wants a separate binary :)

mcfx:
I’ve added a new type and kept it extensible.

However, during local testing I sometimes run into the following issue:
```
Error: Error Domain=VZErrorDomain Code=2 Description="Invalid virtual machine configuration. The storage device attachment is invalid." UserInfo={
    NSLocalizedFailure = "Invalid virtual machine configuration.";
    NSLocalizedFailureReason = "The storage device attachment is invalid.";
}
```

The reproduction is simple, no root privileges required; roughly 1 out of 10 runs may hit the error above.
```
% dd if=/dev/zero of=disk.bin bs=1M count=1024
% hdiutil attach -imagekey diskimage-class=CRawDiskImage -nomount disk.bin
(let the output be /dev/disk4)
% path_to_vfkit --device virtio-blk,path=/dev/disk4,type=dev \
  --cpus 4 \
  --memory 32768 \
  --bootloader efi,variable-store=efi-variable-store,create \
  --device usb-mass-storage,path=archboot-2025.09.22-02.23-6.16.7-1-aarch64-ARCH-local-aarch64.iso,readonly \
  --device virtio-input,keyboard \
  --device virtio-rng \
  --device virtio-net,nat,mac=01:02:03:04:05:06 \
  --device virtio-gpu,width=1920,height=1080 \
  --gui
```

I tried printing the file or attachment obtained by open, but those variables all look fine.
I also tried running two instances of vfkit on the same disk device at the same time, and they both booted normally.

I have no idea how to debug this bug. I’m not sure whether it’s related to other parameters I’m passing to vfkit.

cfergeau:
> Error: Error Domain=VZErrorDomain Code=2 Description="Invalid virtual machine configuration. The storage device attachment is invalid." UserInfo={
>     NSLocalizedFailure = "Invalid virtual machine configuration.";
>     NSLocalizedFailureReason = "The storage device attachment is invalid.";
> }

I’ve seen this error when 2 vfkit instances try to use the same disk image at the same time, could it be that multiple processes are using the disk device when this happens ?

mcfx:
> > Error: Error Domain=VZErrorDomain Code=2 Description="Invalid virtual machine configuration. The storage device attachment is invalid." UserInfo={
> > NSLocalizedFailure = "Invalid virtual machine configuration.";
> > NSLocalizedFailureReason = "The storage device attachment is invalid.";
> > }
> 
> I’ve seen this error when 2 vfkit instances try to use the same disk image at the same time, could it be that multiple processes are using the disk device when this happens ?

That’s actually quite strange. At first I also saw this error occasionally when running it. Since I usually rerun the same command by pressing the “up” key, I suspected some resource might not have been released properly.

To verify, I tried running two instances simultaneously, intentionally using the same disk for read/write — both booted fine, no error.

Then I started testing repeatedly to check how often it fails. It turns out even a completely fresh startup sometimes triggers the same error, so I’m really not sure what’s going on anymore.

pkg/config/virtio.go
```diff
@@ -885,6 +903,8 @@ func (config *DiskStorageConfig) FromOptions(options []option) error {
 		switch option.key {
 		case "path":
 			config.ImagePath = option.value
+		case "type":
+			config.Type = DiskBackendType(option.value)
```
cfergeau: I would check that `option.value` is a valid disk backend, and return an error if not.

pkg/vf/virtio.go
```diff
@@ -87,7 +87,13 @@ func (dev *VirtioBlk) AddToVirtualMachineConfig(vmConfig *VirtualMachineConfigur
 	if err != nil {
 		return err
 	}
-	log.Infof("Adding virtio-blk device (imagePath: %s)", dev.ImagePath)
+	var diskType string
+	if dev.Type == config.DiskBackendDefault {
+		diskType = string(config.DiskBackendImage)
+	} else {
+		diskType = string(dev.Type)
+	}
```
cfergeau, line 98: Same comment here about checking the validity of `dev.Type`.

pkg/vf/virtio.go
```diff
@@ -486,13 +492,50 @@ func AddToVirtualMachineConfig(vmConfig *VirtualMachineConfiguration, dev config
 	}
 }
 
-func (config *DiskStorageConfig) toVz() (vz.StorageDeviceAttachment, error) {
-	if config.ImagePath == "" {
-		return nil, fmt.Errorf("missing mandatory 'path' option for %s device", config.DevName)
+func (conf *DiskStorageConfig) toVz() (vz.StorageDeviceAttachment, error) {
+	if conf.Type == config.DiskBackendImage || conf.Type == config.DiskBackendDefault {
```
cfergeau: I’d use a switch/case here
```
switch conf.Type {
  case config.DiskBackendImage, config.DiskBackendDefault:
    …
  case config.DiskBackendBlockDevice:
    …
  default:
    return nil, fmt.Errorf("unknown disk backend type: %v", conf.Type)
```

You could also add some helpers to make the function shorter (`toVzImageAttachment` and `toVzBlockDeviceAttachment` or such). If you prefer to keep all in `toVz` as it is now, this is fine with me.

pkg/vf/virtio.go
```diff
@@ -486,13 +492,50 @@ func AddToVirtualMachineConfig(vmConfig *VirtualMachineConfiguration, dev config
...
+		} else {
+			open_flags = os.O_RDWR
+		}
+
+		f, err := os.OpenFile(conf.ImagePath, open_flags, 0)
+		if err != nil {
+			return nil, fmt.Errorf("error opening file: %v", err)
+		}
+
+		syncMode := vz.DiskSynchronizationModeFull
```
cfergeau, line 532: Any reason this is different from what is used for disk images? (`		syncMode := vz.DiskImageSynchronizationModeFsync`)

pkg/config/json_test.go
```diff
@@ -289,7 +289,7 @@ var jsonStabilityTests = map[string]jsonStabilityTest{
 		},
 
 		skipFields:   []string{"DevName", "URI"},
-		expectedJSON: `{"kind":"virtioblk","devName":"virtio-blk","imagePath":"ImagePath","readOnly":true,"deviceIdentifier":"DeviceIdentifier"}`,
+		expectedJSON: `{"kind":"virtioblk","devName":"virtio-blk","imagePath":"ImagePath","readOnly":true,"type":"Type","deviceIdentifier":"DeviceIdentifier"}`,
```
vyasgun: `type` should be `image` or `dev` in these test cases. Or we can skip the field altogether.

pkg/vf/virtio.go
mcfx: It's defined in [DiskSynchronizationMode](https://pkg.go.dev/github.com/Code-Hex/vz/v3#DiskSynchronizationMode), and vz has different type for disk and disk image.

Between `DiskSynchronizationModeFull` and `DiskSynchronizationModeNone`, I think the first one is more appropriate.

pkg/vf/virtio.go
mcfx: I've updated these three places.

pkg/config/virtio_test.go
```diff
@@ -105,6 +105,28 @@ func TestVirtioDevices(t *testing.T) {
 			expectedCmdLine:  []string{"--device", fmt.Sprintf("virtio-blk,path=%s,deviceId=test", testImagePath)},
 			alternateCmdLine: []string{"--device", fmt.Sprintf("virtio-blk,deviceId=test,path=%s", testImagePath)},
 		},
+		"NewVirtioBlkWithType": {
```
vyasgun, line 108: Can we add a test for `DiskBackendDefault`?

pkg/vf/virtio.go
```diff
@@ -486,13 +495,51 @@ func AddToVirtualMachineConfig(vmConfig *VirtualMachineConfiguration, dev config
...
+		}
+
+		var open_flags int
+		if conf.ReadOnly {
+			open_flags = os.O_RDONLY
+		} else {
+			open_flags = os.O_RDWR
+		}
+
+		f, err := os.OpenFile(conf.ImagePath, open_flags, 0)
```
vyasgun, line 527: I think `f` should be closed even when there's no error (on VM stop). Should we register an exit handler here?

cfergeau: I think Code-Hex/vz or the virtualization framework take ownership of the file and will close it even if it’s not 100% clear, see https://github.com/Code-Hex/vz/blob/c17cd79c8418115e1fcbf98c78f53bc62fd0c972/storage.go#L380

cfergeau: Following the code through https://github.com/Code-Hex/vz/blob/c17cd79c8418115e1fcbf98c78f53bc62fd0c972/virtualization_14.m#L40 and https://developer.apple.com/documentation/virtualization/vzdiskblockdevicestoragedeviceattachment?language=objc it is really not obvious who will close the file descriptor…
I’m not even sure what will prevent `f` from being garbage collected and closed early (while the attachment is still in use)

cfergeau: Might be related to https://github.com/crc-org/vfkit/pull/373#issuecomment-3377432432 …

pkg/vf/virtio.go
cfergeau: Ah my bad, I did not realize they were different enums, thanks for the clarification!

pkg/config/json_test.go
mcfx: fixed in https://github.com/mcfx/vfkit/commit/0782acf0200130b88791657dcf2326e6197eafe6

pkg/config/virtio_test.go
mcfx: fixed in https://github.com/mcfx/vfkit/commit/9696108d9d5a44d9b4751574af75271dc309fa83

pkg/vf/virtio.go
mcfx: You’re right — that was the cause of the previous error.

I tried manually adding `runtime.GC()` inside this loop:

```go
func (cfg *VirtualMachineConfiguration) toVz() (*vz.VirtualMachineConfiguration, error) {
	for _, dev := range cfg.config.Devices {
		if err := AddToVirtualMachineConfig(cfg, dev); err != nil {
			return nil, err
		}
	}
```

and it consistently triggered the error.

Then I tested https://github.com/mcfx/vfkit/commit/235103f7e0318fbcc8c9c8a030cef9fb91c30131, and it worked — no errors anymore.

I’ve already committed the version that registers the exit handler.

vyasgun: @mcfx I tried to reproduce this error but wasn't able to with the older changes. I am using the same command in [this comment](https://github.com/crc-org/vfkit/pull/373#issuecomment-3377432432).

cfergeau: I’ve created https://github.com/Code-Hex/vz/issues/201

mcfx: @vyasgun I can’t always trigger this bug either. I switched to the commit from that time and ran it hundreds of times in a row without hitting the bug. I guess factors like system load might also affect the timing of GC, which makes it impossible for me to reproduce the exact same environment, so I manually invoked runtime.GC() to trigger it.

PR #375, opened 2025-09-30
ignition: Don’t use a unix socket for the HTTP server

cfergeau:
The ignition code starts an HTTP server used by the VM to fetch its
config. This HTTP server currently listens on a unix socket.
This is a bit wasteful since this runs in the vfkit process, we could
directly use a vsock listener. This also sidesteps another issue which
is that the path to the unix socket is hardcoded and can’t be changed.


lstocchi (commented):
/lgtm

cfergeau:
coderabbit has a good point:
> Current code can append two vsock devices, violating the “0 or 1” constraint and causing config validation to fail when both features are enabled.

Need to double check this

cfergeau:
> coderabbit has a good point:
> 
> > Current code can append two vsock devices, violating the “0 or 1” constraint and causing config validation to fail when both features are enabled.
> 
> Need to double check this

`func (dev *VirtioVsock) AddToVirtualMachineConfig(vmConfig *VirtualMachineConfiguration)` returns early if there is already a vsock device in the vm so this should be fine.

vyasgun:
Looks good to me.

PR #461, opened 2026-06-16
cloud-init: clean up orphaned temp ISOs on startup

vyasgun:
Remove stale cloud-init ISO files left after abrupt termination. Use a dedicated temp directory with PID-tagged filenames so cleanup does not remove ISOs from other running vfkit instances.

## Release Notes

* **Bug Fixes**
  * Prevent stale temporary cloud-init ISO files from accumulating by cleaning up older ISOs tied to inactive PIDs during startup.
  * Improve shutdown reliability by ensuring exit-handler mutex locking is always released safely, even if a handler panics.

* **New Features**
  * Centralize temporary cloud-init ISO creation with a shared utility that uses a dedicated temp subdirectory and PID-based filenames.

### DIFF
--- a/cmd/vfkit/main.go
+++ b/cmd/vfkit/main.go
@@ -406,9 +406,9 @@ func createCloudInitISO(files map[string]io.Reader) (string, error) {
 		}
 	}
 
-	isoFile, err := os.CreateTemp("", "vfkit-cloudinit-")
+	isoFile, err := util.CreateCloudInitISOFile()
 	if err != nil {
-		return "", fmt.Errorf("unable to create temporary cloud-init ISO file: %w", err)
+		return "", err
 	}
 
 	defer func() {
--- a/pkg/util/credtemp.go
+++ b/pkg/util/credtemp.go
@@ -1,0 +1,58 @@
+package util
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/shirou/gopsutil/v4/process"
+	"github.com/sirupsen/logrus"
+)
+
+const cloudInitPrefix = "vfkit-cloudinit"
+
+var cloudInitTempDir = filepath.Join(os.TempDir(), cloudInitPrefix)
+
+func CreateCloudInitISOFile() (*os.File, error) {
+	if err := os.MkdirAll(cloudInitTempDir, 0700); err != nil {
+		return nil, fmt.Errorf("unable to create directory %s: %w", cloudInitTempDir, err)
+	}
+	pid := os.Getpid()
+	file, err := os.CreateTemp(cloudInitTempDir, fmt.Sprintf("%s-%d-*.iso", cloudInitPrefix, pid))
+	if err != nil {
+		return nil, fmt.Errorf("unable to create cloud-init ISO temporary file: %w", err)
+	}
+	return file, nil
+}
+
+func CleanupStaleCloudInitISO() error {
+	entries, err := os.ReadDir(cloudInitTempDir)
+	if os.IsNotExist(err) {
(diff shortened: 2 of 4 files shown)


pkg/util/credtemp.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/shirou/gopsutil/v4/process"
+)
+
+var cloudInitTempDir = filepath.Join(os.TempDir(), "vfkit-cloudinit")
```
baude: vfkit-cloudinit is used quite a bit here? do you think its worth defining it as a const?  total nit obviously

vyasgun: Done - extracted cloudInitPrefix as an unexported const in [credtemp.go](https://github.com/vyasgun/vfkit/blob/3c5480c34dcf701a472b4dc3508f4eba04334a0d/pkg/util/credtemp.go#L13).

pkg/util/credtemp.go
```diff
@@ -0,0 +1,52 @@
...
+	return os.CreateTemp(cloudInitTempDir, fmt.Sprintf("%s-%d-*.iso", cloudInitPrefix, pid))
+}
+
+func CleanupStaleCloudInitISO() error {
+	entries, err := os.ReadDir(cloudInitTempDir)
+	if os.IsNotExist(err) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("unable to read temporary cloud-init ISO file: %w", err)
```
cfergeau: This is a directory, not a file. Maybe you can also replace "temporary cloud-init ISO file" with the value of `cloudInitTempDir`?

pkg/util/credtemp.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+		exists, err := process.PidExists(pid)
+		if err != nil {
+			continue
+		}
+		if exists {
+			continue
+		}
+		if err := os.Remove(filepath.Join(cloudInitTempDir, filename)); err != nil {
+			return fmt.Errorf("unable to remove temporary cloud-init ISO file: %w", err)
```
cfergeau: Same here, the error message could contain the filename.
I’d add a debug log somewhere indicating we removed a file.

cmd/vfkit/main.go
cfergeau, line 422: It’s good that we have this exit handler. Maybe add to the commit log in which scenarios the cleanup on startup is expected to trigger? (ie clarify "abrupt termination" - is this ctrl+c? or other kind of signals?)

cfergeau (commented):
A few minor comments, but /lgtm!

pkg/util/credtemp.go
vyasgun: thanks for pointing out. Changed the message

cmd/vfkit/main.go
vyasgun: I added this to the commit message:

```
Exit handlers run when vfkit shuts down cleanly, including:
      - guest OS stops the VM
      - SIGINT  (Ctrl+C)
      - SIGTERM (kill <pid>)
    In those cases, the VM stops, RunE returns, and defer util.ExecuteExitHandlers()
    removes the cloud-init ISO.
    Startup cleanup removes stale ISOs when exit handlers never run, e.g.:
      - SIGKILL (kill -9 <pid>)
      - process crash/panic
    On the next vfkit start, remove ISOs in the cloud-init temp directory whose
    owning PID is no longer running, without deleting files from other live
    vfkit instances.
```

cfergeau (approved):

PR #491, opened 2026-09-08
boot: allow Linux guests to omit the initrd

Ormod:
Linux boot fails with `invalid initial RAM disk path: stat : no such file or directory` when no initrd is supplied, because the bootloader always applies `WithInitrd("")`.

Construct the options with `WithCommandLine` and append `WithInitrd` only for a nonempty path. Explicitly supplied paths retain their existing validation. The configuration helper also omits `--initrd` when absent, and the usage documentation describes the optional argument.

Regression tests cover omitted, existing, and nonexistent initrd paths, plus command-line generation with and without an initrd.

Validation: `make test-unit`, `make lint`, and `go vet ./...` pass on this branch.

## Summary by CodeRabbit

- **New Features**
  - Linux boot configurations can now omit an initrd and boot without one.
  - When an initrd path is provided, it must point to an existing file.

- **Bug Fixes**
  - Removed failures caused by passing an empty initrd path.
  - Bootloader command-line generation now includes initrd settings only when configured.

- **Documentation**
  - Updated Linux bootloader guidance, including the deprecated `--initrd` option, to describe optional initrd behavior.


vyasgun (changes requested):
I tested this and it is failing without `--initrd`:
```bash
./out/vfkit --cpus 2 --memory 1024 \
  --kernel puipui/Image \
  --kernel-cmdline "quiet" \
  --device virtio-net,nat \
  --device virtio-serial,logFilePath=console.log
Error: if any flags in the group [kernel initrd kernel-cmdline]are set they must all be set; missing [initrd]
```
This is happening because of this line in `pkg/cmdline/cmdline.go`
```bash
cmd.MarkFlagsRequiredTogether("kernel", "initrd", "kernel-cmdline")
```

`--initrd` is no longer required a flag with `--kernel`, this should be:
```bash
cmd.MarkFlagsRequiredTogether("kernel", "kernel-cmdline")
```

